PR #84, opened 2015-03-19
Hello simple

pleonex:
This PR implements the *hello_simple* example from the GettingStarted manual for the new API of C++03. CPPPSM-326.


pleonex:
This is a RTI workspace example, it  will be shipped with next release.

Issue #89, opened 2015-06-26
Improvement of Polling_QueryCondition example


pleonex:
As @alexcamposruiz has suggested me, we could improve this example to show the possibilities of a *querycondition* filter and a more *actual user case*. I propose the following changes:

* Change the *querycondition* parameter in run-time, for instance as we are doing in the *content_filtered_topic* example, we can *rotate* the parameter value after 5 seconds.

* Use a more complex filter expression like `id MATCH %0 AND x > %1`.

* Change the idl type to use an user case like Stock, Flight or similar.

pleonex:
This issue was fixed with #115.

PR #115, opened 2015-07-30
Redesign Polling QueryCondition example

pleonex:
Change of the *polling_querycondition* examples to show all the possibilities of the *QueryCondition* API.
This PR implement the suggestions of #89 in the following way:

* Change IDL type to use a world-real case: share flight information.
* The *QueryCondition* parameters are changed in run-time to get information flights from different companies.
* Use a filter expression with two operations: `MATCH` and `>=`.

The code review for the example in *C* can be started.


pleonex:
C++ and Java examples implemented too.

pleonex:
Implemented in C# too. Pull request finished, code review for all languages can start.

PR #146, opened 2016-04-05
Bug Fixing

menudoproblema:
Error checking if second publisher was created correctly


fgaranda:
Good catch! Thank you Vicente.

PR #154, opened 2017-08-17
Zero copy example

harishkumaruk:
Addition of a new example to demonstrate the usage of zero copy


examples/zero_copy/c++/zero_copy_subscriber.cxx
```diff
@@ -0,0 +1,497 @@
...
+            // Create the frame_set_view with the key and size. This frame set
+            // view attaches to the shared memory segment created by the frame
+            // set managed by the matched DataWriter
+            FrameSetView *frame_set_view = new FrameSetView(key, arg.size);
+            DDS::GUID_t publication_guid;
+            DDS_BuiltinTopicKey_to_guid(
+                    &publication_data.key,
+                    &publication_guid);
+
+            frame_set_views[publication_guid] = frame_set_view;
```
alexcamposruiz: If the guid is duplicate, this will overwrite the existing pointer and leak it. Should check first if the key exists (frame_set_views.find(publication_guid) == frame_set_views.end()).

I guess that will only happen if the event signals that the pub is being disposed (event count change = -1)

examples/zero_copy/c++/zero_copy_subscriber.cxx
```diff
@@ -0,0 +1,497 @@
...
+        std::cout << "Average latency: "
+                << ((total_latency.sec * 1000000)
+                        + (total_latency.nanosec / 1000)) / received_frames
+                << " microseconds" << std::endl;
+    }
+};
+
+void ZeroCopyListener::on_data_available(DataReader* reader)
+{
+    ZeroCopyDataReader *ZeroCopy_reader = NULL;
```
alexcamposruiz: declare variables when they're first used

alexcamposruiz: I know a lot of them are from generated code, but don't follow the same (incorrect) pattern for the rest.

examples/zero_copy/c++/zero_copy_subscriber.cxx
```diff
@@ -0,0 +1,497 @@
...
+
+            // Get the frame_set_view associated with the virtual guid
+            // This is done because there may be multiple Publishers
+            // publishing frames. Each publisher creates its own
+            // shared memory segement to store frames and is identified by
+            // info_seq[i].publication_virtual_guid
+            FrameSetView *frame_set_view = get_frame_set_view(
+                    info_seq[i].publication_virtual_guid);
+
+            if (frame_set_view != NULL) {
```
alexcamposruiz: checking for the opposite condition and calling continue makes the code cleaner.

examples/zero_copy/c++/zero_copy_subscriber.cxx
```diff
@@ -0,0 +1,497 @@
...
+                        0);
+
+                // Validate the computed checksum with the checksum sent by the
+                // DataWriter
+                if (computed_checksum != received_checksum) {
+                    std::cout << "Checksum NOT OK" << std::endl;
+                }
+
+                // Get the current time for computing latency
+                clock->getTime(clock, &current_time);
```
alexcamposruiz: I would extract the time/clock operations into a separate function

examples/zero_copy/c++/zero_copy_subscriber.cxx
harishkumaruk: I investigated this a bit. The on_subscription_matched is called when the pub is disposed but the call would return at get_matched_publication_data() with an error code. Hence we don't reach the point where we create a new FrameSetView.

I have put in a safety check to store the FrameSetView only if there are no existing entries.

harishkumaruk:
Added refactored code based on review comments

examples/zero_copy/c++/zero_copy_publisher.cxx
```diff
@@ -208,23 +203,22 @@ extern "C" int publisher_main(CommandLineArguments arg)
...
+    do {
         writer->get_publication_matched_status(matchedStatus);
 
         if (matchedStatus.total_count == arg.dr_count) {
             break;
         }
-    }
-
-    int index = 0;
+    } while(1);
```
alexcamposruiz: I meant `do { writer->get_publication_matched_status(matchedStatus); } while(matchedStatus.total_count == arg.dr_count);`

examples/zero_copy/c++/zero_copy_subscriber.cxx
```diff
@@ -235,76 +245,75 @@ void ZeroCopyListener::on_data_available(DataReader* reader)
...
+                    0);
+
+            // Validate the computed checksum with the checksum sent by the
+            // DataWriter
+            if (computed_checksum != received_checksum) {
+                std::cout << "Checksum NOT OK" << std::endl;
+            }
+
+            // Get the current time for computing latency
+            get_current_time(clock, &current_time);
```
alexcamposruiz: I meant to encapsulate more than just getting the current time, also packing, computing the diff and unpacking. compute_latency?

examples/zero_copy/c++/zero_copy_subscriber.cxx
```diff
@@ -426,17 +425,19 @@ extern "C" int subscriber_main()
...
+            << std::endl;
+    Duration_t receive_period = {4,0};
+    do {
         if (reader_listener->start_shutdown()) {
             std::cout << "Subscriber application shutting down" << std::endl;
             break;
         }
         NDDSUtility::sleep(receive_period);
-    }
+    } while(1);
```
alexcamposruiz: replace with `while(!reader_listener->start_shutdown()) { NDDSUtility::sleep(receive_period); } ... cout << ...`

harishkumaruk:
Updated code about the while loops, sorry I had mis understood your comment. 
I also refactored a bit the latency computation and use of clock.

alexcamposruiz:
It looks good.

PR #160, opened 2018-12-18
Improve build steps to account for other build generators

titosanxez:


examples/routing_service_shapes_processor/cpp/README.rst
```diff
@@ -19,25 +19,32 @@ From the directory containing the example sources:
...
+- ``<Connext directory>`` shall be replaced with the path to the installation
+  directory of *RTI Connext*.
 - ``<ARCH>`` shall be replaced with the target architecture where you are
   running the example (e.g., x64Darwin15clang7.0).
 
+.. note::
+
+    Certain CMake generators may require specific value for the target platform
+    through the ``-DCMAKE_GENERATOR_PLATFORM``. For example on Windows system,
+    you can specify the values ``x86`` or ``x64``.
```
iblancasa: I think you should add a link to [CMake Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html).

Also, depending on the CMake version, the values for the generators are not the same. For instance:
``cmake -G "Visual Studio 15 2017 Win64"`` is now ``cmake -G "Visual Studio 15 2017" -A x64``

Note ``Win64`` was part of the generator name. Now, `` -A x64`` is a different option for CMake.

examples/routing_service_shapes_processor/cpp/README.rst
```diff
@@ -19,25 +19,32 @@ From the directory containing the example sources:
...
 On successful execution of CMake, you will find different new files in your
-build directory, including a file with name ``Makefile``. This file represents
+build directory that will perform the build of the sources into a shared library.
+
+In Unix, you will find a file with name ``Makefile``. This file represents
 a Unix makefile that contains the rules to build a shared library containing
 the |PROCESSOR| plug-in implementation.
 
-Now run:
+To perform the build run:
```
iblancasa: I think, it would be a better idea to suggest the ``cmake --build .`` command. It doesn't depend on the generator.

If you decide to suggest it, note you need to indicate the configuration for multi-configuration generators (like VS) with the ``--config`` parameter.

iblancasa: Sorry, I don't know what is the ``-A`` parameter?

If you use ``cmake --build .`` you don't need to think about the command to build. CMake will execute the correct one for you.

Regarding the ``--config`` parameter for the multiconfiguration projects, it is needed. If you try to run ``cmake --build .`` without that parameter, CMake will use one by default. For instance, it will use ``Debug`` for Visual Studio Solutions. That will work when CMake was called using ``-DCMAKE_BUILD_TYPE=Debug`` but will fail if CMake was called using ``-DCMAKE_BUILD_TYPE=Release``

titosanxez: You mentioned the "-A x64" in your previous comment,  when calling CMake.
I want to know the behavior of --config if it's not specified when running a multi-configuration generator. Will it fail or will it pick the default (usually VS allows a default configuration).

PR #162, opened 2018-12-29
Port the examples to CMake

iblancasa:
Port the C and C++ examples to use CMake as build system.


iblancasa:
All the C and C++ examples were ported.
How should we document this feature?

fgaranda:
I think to document

> All the C and C++ examples were ported.
> How should we document this feature?

I think we should do a search and replace of the instructions that describe how to build every example (C and C++). They are all part of a section called "Building <ProgrammingLanguage> Example"

fgaranda:
Also, I think we should create a top-level CMake file to build all examples, that is useful some times. The top-level CMake file would include all the projects that you've currently added to build each individual example.

iblancasa:
@fgaranda, I was thinking of replacing the build instructions by this text. What do you think?

I wouldn't like to spend a lot of time changing the readme files.

***
Build the example :wrench:
==========================

All the commands and syntax used assume a Unix-based system. If you run this example in a different architecture, please adapt the commands accordingly.

To build the example you will need first to run CMake to generate the build files. From the directory containing the example sources:

    mkdir build
    cd build
    cmake .. 

CMake parameters
----------------

-   -DCMAKE\_BUILD\_TYPE: specify the build mode. Valid values are `Release` and `Debug`. It is described in the [CMake documentation]
-   -DBUILD\_SHARED\_LIBS: specify the link mode. Valid values are `ON` (for dynamic linking) and `OFF` (for static linking). It is described in the [CMake documentation][1].
-   -G: CMake generator. The generator is the native build system to use build the source code. All the valid values are described described in the CMake documentation [CMake Generators Section].

The CMake scripts will try to guess the values for `CONNEXTDDS_DIR` and `CONNEXTDDS_ARCH`:

-   If you don\'t specify a version or you have installed Connext DDS in a non-default location, you must set the `CONNEXTDDS_DIR` pointing to your RTI Connext DDS installation folder. For example:

        cmake -DCONNEXTDDS_DIR=/home/rti/rti_connext_dds-x.y.z

-   If you have installed more than one architecture on your system (i.e., more than one target rtipkg), you must set the `CONNEXTDDS_ARCH` to provide the name of the architecture. For example:

        cmake -DCONNEXTDDS_ARCH=x64Linux3gcc5.4.0

Run the native tool
-------------------

On successful execution of CMake, you will find different new files in your build directory (these files will depend on the selected CMake Generator). To build the source code, run:

    cmake --build .

**Important**: if you are using a multiconfiguration generator like Visual Studio Solutions, you need to specify the configuration to build. If you are building the project in release mode, you need to build your
code running:

    cmake --build . --configuration Release

If you are building debug mode, you need to run:

    cmake --build . --configuration Debug

  [CMake documentation]: https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html
  [1]: https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html
  [CMake Generators Section]: https://cmake.org/cmake/help/v3.13/manual/cmake-generators.7.html
***

resources/cmake/ConnextDdsAddExample.cmake
```diff
@@ -0,0 +1,276 @@
...
+        set(extension "c")
+        if("${_CONNEXT_LANG}" STREQUAL "C++" OR
+            "${_CONNEXT_LANG}" STREQUAL "C++03" OR
+            "${_CONNEXT_LANG}" STREQUAL "C++11")
+            set(extension "cxx")
+        endif()
+
+        set(publisher_src
+            "${CMAKE_CURRENT_SOURCE_DIR}/${_CONNEXT_IDL}_publisher.${extension}"
+            ${${_CONNEXT_IDL}_${lang_var}_SOURCES}
```
jpovedano: We may need to create a object library with the type sources. Otherwise, we will have two different targets depending on generated files, that will cause problems if we try to use parallel builds

iblancasa: Yes! I solved the issue with my last commit.

Thanks!

examples/connext_dds/asyncwaitset/c++03/README.md
```diff
@@ -1,76 +1,171 @@
...
+    for the types defined in the IDL file associated with the example.
+
+2.  Build the corresponding Publisher and Subscriber applications.
+
+3.  Copy the USER_QOS_PROFILES.xml file into the directory where the publisher
+    and subscriber executables are generated.
+
+You will find the definition of connextdds_add_example, along with detailed
+documentation, in
+[rescources/cmake/ConnextDdsAddExample.cmake](../../../../rescources/cmake/ConnextDdsAddExample.cmake).
```
jpovedano: Typo: "rescources" instead of "resources"

PR #166, opened 2019-01-02
Add Routing Service C++ adapter example

jpovedano:
This PR adds a Routing Service adapter that uses the Modern C++ API. 
This example uses CMake to build the example


examples/routing_service/routing_service_file_adapter/cpp/FileConnection.hpp
```diff
@@ -0,0 +1,69 @@
...
+public:
+
+    FileConnection(
+            StreamReaderListener *input_stream_discovery_listener,
+            StreamReaderListener *output_stream_discovery_listener,
+            const PropertySet &properties) :
+            discovery_reader_(properties)
+    {
+        // Once the FileConnection is initialized, we trigger an event
+        // to notify that the streams are ready.
```
titosanxez: this should be implemented in the .cxx file

examples/routing_service/routing_service_file_adapter/cpp/FileConnection.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+private:
+    FileDiscoveryStreamReader discovery_reader_;
+};
+
+}  // namespace examples
+}  // namespace community
+}  // namespace rti
+
+#endif
```
titosanxez: missing new line

examples/routing_service/routing_service_file_adapter/cpp/FileConnection.hpp
```diff
@@ -0,0 +1,69 @@
...
+    void delete_stream_reader(StreamReader *reader);
+
+    StreamWriter *create_stream_writer(
+            Session *session,
+            const StreamInfo &info,
+            const PropertySet &properties);
+
+    void delete_stream_writer(StreamWriter *writer);
+
+    DiscoveryStreamReader *input_stream_discovery_reader() {
```
titosanxez: implement in .cxx

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,129 @@
...
+    for (const auto &it : property) {
+        if (it.first == INPUT_FILE_PROPERTY_NAME) {
+            inputfile_.open(it.second);
+        } else if (it.first == SAMPLE_PERIOD_PROPERTY_NAME) {
+            sampling_period_ = std::stoi(it.second);
+        }
+    }
+
+    if (!inputfile_.is_open()) {
+        throw std::logic_error("Input file not provided or unable to open");
```
titosanxez: I believe we should be using dds::core::error. Do you gree @alexcamposruiz ?

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,129 @@
...
+
+using namespace rti::community::examples;
+using namespace dds::core::xtypes;
+
+const std::string FileStreamReader::INPUT_FILE_PROPERTY_NAME = "example.adapter.input_file";
+const std::string FileStreamReader::SAMPLE_PERIOD_PROPERTY_NAME = "example.adapter.sample_period_sec";
+
+void FileStreamReader::ProcessThread()
+{
+    stop_thread_ = false;
```
titosanxez: this should be initialized in the constructor

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    filereader_thread_ = std::thread(&FileStreamReader::ProcessThread, this);
+}
+
+void FileStreamReader::read(
+        std::vector<dds::core::xtypes::DynamicData *> &samples,
+        std::vector<dds::sub::SampleInfo *> &infos)
+{
+    take(samples, infos);
```
titosanxez: implementing read is not required. Default implementation will throw an exception. So if we don't intentd to implement actually the read, we should leave with default implementation. For basic forwarding (when default processor is working), only take(data, info) is required.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.hpp
```diff
@@ -0,0 +1,82 @@
...
+    void take(
+            std::vector<dds::core::xtypes::DynamicData *> &,
+            std::vector<dds::sub::SampleInfo *> &,
+            const SelectorState &selector_state);
+
+    void return_loan(
+            std::vector<dds::core::xtypes::DynamicData *> &,
+            std::vector<dds::sub::SampleInfo *> &);
+
+    ~FileStreamReader()
```
titosanxez: implement in .cxx file

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.hpp
```diff
@@ -0,0 +1,82 @@
...
+        filereader_thread_.join();
+        inputfile_.close();
+    }
+
+private:
+
+    static const std::string INPUT_FILE_PROPERTY_NAME;
+    static const std::string SAMPLE_PERIOD_PROPERTY_NAME;
+
+    void ProcessThread();
```
titosanxez: naming convention?

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
```diff
@@ -0,0 +1,52 @@
...
+#include <rti/topic/to_string.hpp>
+
+using namespace rti::community::examples;
+
+const std::string FileStreamWriter::OUTPUT_FILE_PROPERTY_NAME = "example.adapter.output_file";
+
+FileStreamWriter::FileStreamWriter(const PropertySet &property)
+{
+
+    // const auto &it = property.find(OUTPUT_FILE_PROPERTY_NAME);
```
titosanxez: remove leftover comments

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+class FileStreamWriter : public DynamicDataStreamWriter {
+public:
+    FileStreamWriter(const PropertySet &);
+
+    int write(
+            const std::vector<dds::core::xtypes::DynamicData *> &samples,
+            const std::vector<dds::sub::SampleInfo *> &infos);
+
+    ~FileStreamWriter()
```
titosanxez: implement in .cxx

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.hpp
```diff
@@ -0,0 +1,48 @@
...
+    {
+        outputfile_.close();
+    };
+
+private:
+
+    static const std::string OUTPUT_FILE_PROPERTY_NAME;
+
```
titosanxez: remove line

examples/routing_service/routing_service_file_adapter/cpp/README.md
```diff
@@ -0,0 +1,79 @@
...
+build it into a shared library and load it with RoutingService.
+
+In this concrete example, we show how a simple adapter to import/export data from/to files using RTI Routing Service.
+
+The code in this directory provides the following components:
+
+- `FileAdapter` implements the plugin that is loaded by *RTI Routing Service*. It is responsible to create and delete connections.
+- `FileConnection` implements a connection. This component is responsible of the creation and deletion of `StreamReaders` and `StreamWriters`.
+- `FileDiscoveryStreamReader` implements the logic necessary to propagate information about newly discovered streams to RTI Routing Service.
+- `FileStreamReader` implements an `StreamReader` that reads sample information from a CSV file.
```
titosanxez: why reading from CSV and writing to JSON? should we be symmetric? or even better, have an option to decide the format?

examples/routing_service/routing_service_file_adapter/cpp/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+Received Sample (JSON):
+{
+   "id":1,
+   "message":" Hello World"
+}
+
+```
+
+The general behavior of this example, can be controlled using properties. You
```
titosanxez: remove command in the middle of the sentence

examples/routing_service/routing_service_file_adapter/cpp/README.md
```diff
@@ -0,0 +1,79 @@
...
+| `example.adapter.output_file`       | `<output>` | Path to the file where to store the received samples                                           |
+
+
+## Requirements
+
+To run this example you will need:
+
+- RTI Connext Professional version 6.0.0 or higher.
+- CMake version 3.10 or higher
+- A target platform supported by RTI RoutingService.
```
titosanxez: we need to add 'with support for C++11"

examples/routing_service/routing_service_file_adapter/cpp/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+In order to build this example, you need to provide the following variables to
+`CMake`:
+
+- `CONNEXTDDS_DIR`
+- `CONNEXTDDS_ARCH`
+
+```bash
+mkdir build
+cmake -DCONNEXTDDS_DIR=<connext dir> -DCONNEXTDDS_ARCH=<connext architecture> ..
```
titosanxez: do we need to add -DBUILD_SHARED_LIBS=ON? or is that default?

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+add_library(
+    fileadapter
+    MODULE
+        ${CMAKE_SOURCE_DIR}/FileAdapter.cxx
+        ${CMAKE_SOURCE_DIR}/FileConnection.cxx
+        ${CMAKE_SOURCE_DIR}/FileDiscoveryStreamReader.cxx
+        ${CMAKE_SOURCE_DIR}/FileStreamReader.cxx
+        ${CMAKE_SOURCE_DIR}/FileStreamWriter.cxx
+)
+
```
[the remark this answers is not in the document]
jpovedano: No, they are not needed. Here we are using virtual targets, and the virtual targets have in their PUBLIC interface the appropriate include directories and compile definitions.

When using virtual targets, they are propagated automatically by CMake

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ find_package(
 # Define the library that will provide the adapter plugin
 add_library(
     fileadapter
-    MODULE
+    SHARED
```
iblancasa: Why did you change this?

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
```diff
@@ -45,6 +45,19 @@ target_link_libraries(
 
 # Copy the sample data file into the binary directory
 configure_file(
```
iblancasa: Actually we should be using a `custom_command` to copy the files

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
pralhadsapre: Cause it wasn't working on MacOS. It was creating the library as libfileadapter.so. Whereas we were looking for libfileadapter.dylib.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
pralhadsapre: Is there a reason behind this preference? Any best practices?

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.hpp
pralhadsapre: Done

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.hpp
pralhadsapre: Done

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.hpp
pralhadsapre: Changed to file_reading_thread()

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.hpp
pralhadsapre: Done

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
pralhadsapre: Done. Initialized in the initializer list.

examples/routing_service/routing_service_file_adapter/cpp/FileConnection.hpp
pralhadsapre: Done

examples/routing_service/routing_service_file_adapter/cpp/README.md
pralhadsapre: We now read from and write in a CSV format for ShapeType.

examples/routing_service/routing_service_file_adapter/cpp/README.md
pralhadsapre: Done

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.cxx
```diff
@@ -0,0 +1,33 @@
...
+    FileConnection *fc = new FileConnection(
+            input_stream_discovery_listener,
+            output_stream_discovery_listener,
+            properties);
+    return fc;
+};
+
+rti::config::LibraryVersion FileAdapter::get_version() const
+{
+    return rti::config::LibraryVersion(6, 0, 0, 'r');
```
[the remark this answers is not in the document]
titosanxez: @pralhadsapre can you set this to 1.0.0?

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.hpp
```diff
@@ -0,0 +1,54 @@
...
+#ifndef FILEADAPTER_HPP
+#define FILEADAPTER_HPP
+
+#include <dds/core/corefwd.hpp>
+#include <dds/core/macros.hpp>
+#include <rti/routing/PropertySet.hpp>
+#include <rti/routing/RoutingService.hpp>
+#include <rti/routing/adapter/AdapterPlugin.hpp>
+
+using namespace rti::routing;
```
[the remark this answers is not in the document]
titosanxez: @pralhadsapre  can you fix this?

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.hpp
```diff
@@ -0,0 +1,54 @@
...
+using namespace rti::routing::adapter;
+
+namespace rti {
+namespace community {
+namespace examples {
```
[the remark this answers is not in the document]
titosanxez: @pralhadsapre  can you check on this?

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.hpp
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * The generated symbol has the name:
+ *
+ * \code
+ * FileAdapterPlugin_create_adapter_plugin
+ * \endcode
+ */
+RTI_ADAPTER_PLUGIN_CREATE_FUNCTION_DECL(FileAdapter)
+
+#endif
```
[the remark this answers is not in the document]
titosanxez: @pralhadsapre

examples/routing_service/routing_service_file_adapter/cpp/FileConnection.cxx
```diff
@@ -0,0 +1,77 @@
...
+        StreamReaderListener *output_stream_discovery_listener,
+        const PropertySet &properties) :
+        input_discovery_reader_(properties)
+{
+    /** 
+     * Once the FileConnection is initialized, we trigger an event 
+     * to notify that the streams are ready.
+     */
+    input_stream_discovery_listener_ = input_stream_discovery_listener;
+    input_stream_discovery_listener_->on_data_available(&input_discovery_reader_);
```
titosanxez: @pralhadsapre why is this needed?

examples/routing_service/routing_service_file_adapter/cpp/FileConnection.hpp
titosanxez: @pralhadsapre

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
```diff
@@ -0,0 +1,65 @@
...
+    if (fexists(CIRCLE_FILE_NAME)) {
+        this->data_samples_.push_back(new StreamInfo("Circle", "ShapeType"));
+    }
+
+    if (fexists(TRIANGLE_FILE_NAME)) {
+        this->data_samples_.push_back(new StreamInfo("Triangle", "ShapeType"));
+    }
+}
+
+void FileInputDiscoveryStreamReader::dispose()
```
titosanxez: This operation should receive a stream name as an argument, since streams can be disposed independently

titosanxez: Also this class should be the one holding the listener, and notifiying when appropriate

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.hpp
```diff
@@ -0,0 +1,54 @@
...
+    static const std::string TRIANGLE_FILE_NAME;
+
+    std::vector<rti::routing::StreamInfo *> data_samples_;
+};
+
+}  // namespace examples
+}  // namespace community
+}  // namespace rti
+
+#endif
```
titosanxez: missing end line

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
titosanxez: @pralhadsapre use dds::core::IllegalArgument or whatever is the actual exception

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,141 @@
...
+
+using namespace rti::community::examples;
+using namespace dds::core::xtypes;
+
+const std::string FileStreamReader::INPUT_FILE_PROPERTY_NAME = "example.adapter.input_file";
+const std::string FileStreamReader::SAMPLE_PERIOD_PROPERTY_NAME = "example.adapter.sample_period_sec";
+
+void FileStreamReader::file_reading_thread()
+{
+    while (!stop_thread_) {
```
titosanxez: @pralhadsapre this logic is not consistent with the model: this provides the reading of a single file, however the model is that we should be able to read from multiple files at the same time

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
pralhadsapre: Agreed. I have changed it to receive a pointer to StreamInfo.

pralhadsapre: I have also moved the listener inside this class.

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.hpp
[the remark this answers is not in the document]
fgaranda: I think clang-format is configured to format namespaces like that. Maybe we should tweak the script to use the preferred form.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
iblancasa: I think this post can be useful: https://dev.to/iblancasa/learning-cmake-3-understanding-addcustomcommand-and-addcustomtarget-43gp

The reason is that `configure_file` will be run during the build system creation. Probably, it will make more sense to do the copy of the file during the build. But in this case, is up to you :) .

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
titosanxez: Also we support static linking of adapters. It's better to remove the SHARED and indicate in the instructions whether shared libs is on or off

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
```diff
@@ -0,0 +1,110 @@
...
+
+    /** 
+     * Once the FileInputDiscoveryStreamReader is initialized, we trigger an 
+     * event to notify that the streams are ready.
+     */
+    input_stream_discovery_listener_->on_data_available(this);
+}
+
+void FileInputDiscoveryStreamReader::dispose(
+        const rti::routing::StreamInfo *stream_info)
```
titosanxez: pass as reference

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
```diff
@@ -0,0 +1,110 @@
...
+{
+    /**
+     * This guard is essential since the take() and return_loan() operations triggered 
+     * by calling on_data_available() execute on a different thread. The take() and 
+     * return_loan() operations also need to access the data_samples_ list.
+     */
+    std::lock_guard<std::mutex> guard(data_samples_mutex_);
+
+    for (int i = 0; i < stream.size(); i++) {
+        this->data_samples_.pop_front();
```
titosanxez: we need to call delete on the sample, otherwise this is leaking

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,144 @@
...
+FileStreamReader::FileStreamReader(
+        FileConnection *connection, 
+        const StreamInfo &info,
+        const PropertySet &property,
+        StreamReaderListener *listener)
+        : filereader_thread_(),
+        sampling_period_(1), 
+        stop_thread_(false)
+{
+    stream_info_ = new StreamInfo(
```
titosanxez: this one we don't need to make a heap copy; make it a stack member

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+        ${CMAKE_SOURCE_DIR}/FileInputDiscoveryStreamReader.cxx
+        ${CMAKE_SOURCE_DIR}/FileStreamReader.cxx
+        ${CMAKE_SOURCE_DIR}/FileStreamWriter.cxx
+)
+
+target_link_libraries(
+    fileadapter
+    RTIConnextDDS::routing_service_infrastructure
+    RTIConnextDDS::cpp2_api
+    RTIConnextDDS::c_api
```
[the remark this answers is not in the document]
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
...
+# Define the library that will provide the adapter plugin
+add_library(
+    fileadapter
+    SHARED
+        ${CMAKE_SOURCE_DIR}/FileAdapter.cxx
+        ${CMAKE_SOURCE_DIR}/FileConnection.cxx
+        ${CMAKE_SOURCE_DIR}/FileInputDiscoveryStreamReader.cxx
+        ${CMAKE_SOURCE_DIR}/FileStreamReader.cxx
+        ${CMAKE_SOURCE_DIR}/FileStreamWriter.cxx
+)
```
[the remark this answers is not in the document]
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.hpp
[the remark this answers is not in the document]
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileConnection.hpp
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
```diff
@@ -0,0 +1,110 @@
...
+        std::vector<rti::routing::StreamInfo *> &stream)
+{
+    /**
+     * This guard is essential since the take() and return_loan() operations triggered 
+     * by calling on_data_available() execute on a different thread. The take() and 
+     * return_loan() operations also need to access the data_samples_ list.
+     */
+    std::lock_guard<std::mutex> guard(data_samples_mutex_);
+
+    auto iter = data_samples_.begin();
```
alexcamposruiz: This whole loop should be just:

```
std::copy(data_samples_.begin(), data_samples_.end(), std::back_inserter(stream));
```

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
pralhadsapre: I can't since its used in the file_reading_thread()

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
pralhadsapre: Done. Used dds::core::IllegalOperationError to throw.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
titosanxez: why is that a problem? that shouldn't make any difference.

pralhadsapre: Changed it to be a class member instead of a pointer.

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.cxx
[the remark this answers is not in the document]
pralhadsapre: Done

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
pralhadsapre: Done

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
pralhadsapre: Done. Also added additional content in the README.md to describe how the static libraries can be used with RS as a library.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+    RTIConnextDDS::cpp2_api
+    RTIConnextDDS::c_api
+    ${CONNEXTDDS_EXTERNAL_LIBS}
+)
+
+# Copy the sample data file into the binary directory
+configure_file(
+    ${CMAKE_SOURCE_DIR}/samples_in.csv
+    ${CMAKE_BINARY_DIR}/samples_in.csv
+)
```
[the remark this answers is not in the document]
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.hpp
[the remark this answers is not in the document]
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileConnection.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <rti/routing/adapter/Connection.hpp>
+
+#include "FileDiscoveryStreamReader.hpp"
+
+namespace rti {
+namespace community {
+namespace examples {
+using namespace rti::routing;
+using namespace rti::routing::adapter;
+using namespace rti::routing::adapter::detail;
```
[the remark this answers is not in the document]
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
pralhadsapre: Hey @iblancasa good job with the article. That was an interesting read. However this example doesn't require all the intricacies you described. Sin the CSV and XML files are typically immutable, we should be okay with configure_file().

pralhadsapre:
Hello reviewers,

I have made all the changes you requested. Please review the latest commits and let me know if you have any more suggestions.

Regards,
Pralhad

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.cxx
```diff
@@ -0,0 +1,35 @@
...
+    FileConnection *fc = new FileConnection(
+            input_stream_discovery_listener,
+            output_stream_discovery_listener,
+            properties);
+    return fc;
+};
+
+rti::config::LibraryVersion FileAdapter::get_version() const
+{
+    return rti::config::LibraryVersion(1, 0, 0, 'r');
```
alexcamposruiz: Prefer:

```
return {1, 0, 0, 'r'};
```

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <dds/core/corefwd.hpp>
+#include <dds/core/macros.hpp>
+#include <rti/routing/PropertySet.hpp>
+#include <rti/routing/RoutingService.hpp>
+#include <rti/routing/adapter/AdapterPlugin.hpp>
+
+namespace rti { namespace community { namespace examples {
+class FileAdapter : public rti::routing::adapter::AdapterPlugin {
+public:
+    FileAdapter(rti::routing::PropertySet &){};
```
alexcamposruiz: Make it `explicit`.

[C.46: By default, declare single-argument constructors explicit](http://intranet:8000//local/preship/cpp_core_guidelines/sphinx/_build/html/CppCoreGuidelines.html#Rc-explicit)

Remove the trailing `;`

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
```diff
@@ -0,0 +1,60 @@
...
+using namespace rti::routing;
+using namespace rti::routing::adapter;
+using namespace rti::community::examples;
+
+const std::string FileStreamWriter::OUTPUT_FILE_PROPERTY_NAME = "example.adapter.output_file";
+
+FileStreamWriter::FileStreamWriter(const PropertySet &property)
+{
+    std::string output_file_name;
+    for (const auto &it: property) {
```
alexcamposruiz: Don't call this variable `it`. It's not an iterator. I'd call it `p`

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
```diff
@@ -0,0 +1,60 @@
...
+        if (it.first == OUTPUT_FILE_PROPERTY_NAME) {
+            output_file_name = it.second;
+            output_file_.open(output_file_name);
```
alexcamposruiz: Should you break after finding the first matching property?

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
```diff
@@ -0,0 +1,60 @@
...
+        }
+    }
+
+    if (!output_file_.is_open()) {
+        throw dds::core::IllegalOperationError(
+                "Open output file: " + output_file_name);
```
alexcamposruiz: Error should be "Error opening output file..." or "Failed to open output file..."

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
```diff
@@ -0,0 +1,60 @@
...
+    } else {
+        std::cout << "Output file name: " << output_file_name << std::endl;
+    }
+};
+
+int FileStreamWriter::write(
+        const std::vector<dds::core::xtypes::DynamicData *> &samples,
+        const std::vector<dds::sub::SampleInfo *> &infos)
+{
+    for (std::vector<dds::core::xtypes::DynamicData *>::const_iterator it = samples.begin();
```
alexcamposruiz: why not `for (auto sample : samples)`?

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
```diff
@@ -0,0 +1,60 @@
...
+        const std::vector<dds::core::xtypes::DynamicData *> &samples,
+        const std::vector<dds::sub::SampleInfo *> &infos)
+{
+    for (std::vector<dds::core::xtypes::DynamicData *>::const_iterator it = samples.begin(); 
+            it != samples.end(); 
+            ++it) {
+        std::string str = rti::topic::to_string(
+                (**it), 
+                rti::topic::PrintFormatProperty::Default());
+        std::cout << "Received Sample: " << std::endl << str << std::endl;
```
alexcamposruiz: Did you choose to call `to_string()` because you prefer that format? Otherwise `cout << sample` just works.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
```diff
@@ -0,0 +1,60 @@
...
+
+        output_file_ << (*it)->value<std::string>("color") << ","
```
alexcamposruiz: Do you ensure that the data you receive here is of ShapeType (with those 4 fields)?

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,146 @@
...
+     * This is the hardcoded type information about ShapeType.
+     * You are advised to change this as per your type definition
+     */
+    sample->value("color", color);
+    sample->value("x", std::stoi(x));
+    sample->value("y", std::stoi(y));
+    sample->value("shapesize", std::stoi(shapesize));
+
+    dds::sub::SampleInfo *info = new dds::sub::SampleInfo();
+    (*info)->native().valid_data = DDS_BOOLEAN_TRUE;
```
alexcamposruiz: Is it necessary to create a SampleInfo object just to set the valid_data flag to true? Shouldn't it work if you leave the info sequence empty or set the elements to null?

examples/routing_service/routing_service_file_adapter/cpp/FileConnection.cxx
```diff
@@ -0,0 +1,73 @@
...
+};
+
+void FileConnection::delete_stream_writer(StreamWriter *writer)
+{
+    delete writer;
+}
+
+DiscoveryStreamReader *FileConnection::output_stream_discovery_reader() 
+{
+    return NULL;
```
alexcamposruiz: ```return nullptr;```

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
```diff
@@ -0,0 +1,108 @@
...
+{
+    std::ifstream input_file;
+    input_file.open(filename);
+    return input_file.is_open();
+}
+
+FileInputDiscoveryStreamReader::FileInputDiscoveryStreamReader( 
+        const PropertySet &, 
+        StreamReaderListener *input_stream_discovery_listener)
+        : data_samples_mutex_(),
```
alexcamposruiz: Don't have call the default constructor in the initializer list. The default constructor is always called for all members unless you specify a *different* constructor in the initializer list.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
```diff
@@ -0,0 +1,108 @@
...
+     */
+    std::lock_guard<std::mutex> guard(data_samples_mutex_);
+    std::copy(data_samples_.begin(), data_samples_.end(), std::back_inserter(stream));
+}
+
+void FileInputDiscoveryStreamReader::return_loan(
+        std::vector<rti::routing::StreamInfo *> &stream)
+{
+    /**
+     * This guard is essential since the take() and return_loan() operations triggered
```
alexcamposruiz: Why do take() and return_loan() run on different threads? Shouldn't RS call:

take()
...
return_loan()

In the same thread?

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /** 
+     * Once the FileInputDiscoveryStreamReader is initialized, we trigger an 
+     * event to notify that the streams are ready.
+     */
+    input_stream_discovery_listener_->on_data_available(this);
+}
+
+void FileInputDiscoveryStreamReader::dispose(
```
alexcamposruiz: You should explain the purpose of this function.

In general, being the main RS adapter example, I think all key functions need more documentation, explaining how and why RS calls them.

examples/routing_service/routing_service_file_adapter/cpp/.gitignore
```diff
@@ -0,0 +1,3 @@
+.vscode/
+build/
+*.o
```
lulivi: Should we maybe add these patterns to our global `.gitignore` instead of creating a new one in this specific directory?

iblancasa:
Please, fix the issues reported by Travis

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+#  RTI grants Licensee a license to use, modify, compile, and create derivative
+#  works of the Software.  Licensee has the right to distribute object form
+#  only for use with RTI products.  The Software is provided "as is", with no
+#  warranty of any type, including any warranty for fitness for any purpose.
+#  RTI is under no obligation to maintain or support the Software.  RTI shall
+#  not be liable for any incidental or consequential damages arising out of the
+#  use or inability to use the software.
+#
+cmake_minimum_required(VERSION 3.11)
+set (CMAKE_CXX_STANDARD 11)
```
iblancasa: Please, use the [CXX_STANDARD](https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html) property instead.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CONNEXTDDS_DIR}/resource/cmake")
+find_package(
+    RTIConnextDDS "6.0.0"
+    REQUIRED
+    COMPONENTS
+        core
+        routing_service
+)
+
+# Define the library that will provide the adapter plugin
```
iblancasa: Probably this comment is not needed

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+add_library(
+    fileadapter
+    # The kind of library created depends on value of BUILD_SHARED_LIBS=ON|OFF
```
iblancasa: I think this comment is not needed

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+    rti::routing::adapter::Connection *create_connection(
```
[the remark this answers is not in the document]
pralhadsapre: Done for all header files.

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.hpp
pralhadsapre: Done. I removed the {} instead, since this is just a declaration.

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.cxx
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
pralhadsapre: Yes, you are right. I added the break.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
pralhadsapre: Good point. Used cout << *sample as suggested.

pralhadsapre: I tried using cout << *sample. Its a little more indented to the right TBH. I think the to_string() formats the output better.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
pralhadsapre: Yes it will always have those fields. We only deal with the ShapeType in this example.

lulivi:
Please, pull changes from develop branch, there are some Travis CI fixes.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
pralhadsapre: IDK if the valid_data flag is set to true by default. Even if it is so, it would be helpful to demonstrate the use, in case someone wants to use the flag to indicate invalid data.

examples/routing_service/routing_service_file_adapter/cpp/FileConnection.cxx
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
pralhadsapre: Yes it does. Maybe my comment didn't convey the idea clearly. I meant dispose() doesn't run on the same thread as take() or return_loan(). I have updated the comment to be more clear.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
titosanxez: Actually returning SampleInfo is not needed, so we can just leave the info_seq empty.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
titosanxez: take() and return_loan() runs in the same thread, but the function that adds sample to the queue is in a separate thread (The one that reads the files). So that's why the mutex is needed

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
alexcamposruiz: You don't need to pass `Default()`. I would just do this:

```
std::cout << "Received Sample:\n" << rti::topic::to_string(*sample) << std::endl;
```

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.hpp
alexcamposruiz: You can't remove the `{}` can you? Won't RS call that constructor? If you don't need it at all, delete the whole constructor, don't declare it and not define it.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
pralhadsapre: You mean set it per target? Although in this case we just have "fileadapter".

What do you think about adding set(CMAKE_CXX_STANDARD_REQUIRED ON) as well?

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
pralhadsapre: I know this is verbose and redundant. But if the customer isn't well versed with CMake, its better to say more.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
pralhadsapre: Same as below.

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.hpp
pralhadsapre: You are right. I ended up defining and empty constructor body. We can remove that if needed.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamWriter.cxx
pralhadsapre: I like that idea. Done.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
pralhadsapre: I added comments about its use in the .hpp file. For other functions requiring more clarification I will review it with @titosanxez.

examples/routing_service/routing_service_file_adapter/cpp/.gitignore
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
pralhadsapre: Also I am interested to know if there is any specific reason for not using CMAKE_CXX_STANDARD variable.

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.cxx
```diff
@@ -0,0 +1,47 @@
...
+    FileConnection *fc = new FileConnection(
+            input_stream_discovery_listener,
+            output_stream_discovery_listener,
+            properties);
+    return fc;
+};
+
+void FileAdapter::delete_connection(Connection *connection)
+{
+    /**
```
titosanxez: you ned to call delete connection here, otherwise the connection object is leaking

examples/routing_service/routing_service_file_adapter/cpp/FileConnection.cxx
```diff
@@ -0,0 +1,73 @@
...
+{
+    return new FileStreamWriter(properties);
+};
+
+void FileConnection::delete_stream_writer(StreamWriter *writer)
+{
+    delete writer;
+}
+
+DiscoveryStreamReader *FileConnection::output_stream_discovery_reader()
```
titosanxez: I'm not sure if we need to implement this one, since it shouldn't be a pure virtual method. In such case, let's not define it nor declare it

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,149 @@
...
+    std::cout << "Reached end of stream for file: " << input_file_name_ << std::endl;
+    file_connection_->dispose_discovery_stream(stream_info_);
+}
+
+FileStreamReader::FileStreamReader(
+        FileConnection *connection, 
+        const StreamInfo &info,
+        const PropertySet &properties,
+        StreamReaderListener *listener)
+        : filereader_thread_(),
```
titosanxez: no need to call default constructor

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,149 @@
...
+    std::string y;
+    std::string shapesize;
+
+    // Reading input from the csv file
+    std::getline(s, color, ',');
+    std::getline(s, x, ',');
+    std::getline(s, y, ',');
+    std::getline(s, shapesize, ',');
+
+    samples.resize(1);
```
titosanxez: should we return more samples in one read? also, adding a comment indicating the intent is benefitial

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
pralhadsapre: We only read a single line in the file_reading_thread() method.

examples/routing_service/routing_service_file_adapter/cpp/FileConnection.cxx
pralhadsapre: We can discuss this tomorrow. I would want to keep it for the purpose of having skeleton code if a customer is writing their own adapter.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
iblancasa: I don't see the point of setting `CMAKE_CXX_STANDARD_REQUIRED` to `ON`.

The reason to not use `CMAKE_CXX_STANDARD` is simple: you are forcing the full project to use that CXX STANDARD. If this project is added as part of a bigger project (as we did for the ConnextDDS examples), you will be forcing the parent project. It is not a good idea to set things globally if they can be set per target.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
iblancasa: I think it is not a good idea because, if we follow that philosophy we will need to comment each single line.

iblancasa:
Travis is still failing.

examples/routing_service/routing_service_file_adapter/cpp/FileAdapter.cxx
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
pralhadsapre: Ok.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
pralhadsapre: Got it. Thanks for the clarification Israel.

I used the CXX_STANDARD and CXX_STANDARD_REQUIRED properties. The reason for using the latter is we don't want the decay behavior for the compiler flags pertaining to C++11. This should make the CMake generate step fail instead of the compilation.

examples/routing_service/routing_service_file_adapter/cpp/CMakeLists.txt
titosanxez: Pralhad I see your intent of trying to have a good example to guide users but it's true we can't comment on all the cmake basic concepts. What you can do though, potentially, is to add some Note in the readme indicating that this example is prepared to build a dynamic library but that static linking is also supported if you link RS statically into your application.

pralhadsapre: Yes. There is already a mention of that in the README.md. This one was superfluous.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
```diff
@@ -0,0 +1,118 @@
...
+{
+    std::ifstream input_file;
+    input_file.open(filename);
+    return input_file.is_open();
+}
+
+FileInputDiscoveryStreamReader::FileInputDiscoveryStreamReader(
+        const PropertySet &,
+        StreamReaderListener *input_stream_discovery_listener)
+{
```
alexcamposruiz: The constructor is not exception safe. Any failure in push_back or new leaks the previously new'd StreamInfos.

You can make `data_samples_` a `vector<unique_ptr<StreamInfo>>`. That would also simplify the reutnr_loan function to just `v.clear()`.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * This guard is essential since the take() and return_loan() operations
+     * triggered by calling on_data_available() execute on an internal Routing
+     * Service thread. The custom dispose() operation doesn't run on that
+     * thread. Since the take() and return_loan() operations also need to access
+     * the data_samples_ list this protection is required.
+     */
+    std::lock_guard<std::mutex> guard(data_samples_mutex_);
+
+    auto iter = data_samples_.begin();
```
alexcamposruiz: See my comment in the constructor. This loop would be unnecessary.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * This guard is essential since the take() and return_loan() operations
+     * triggered by calling on_data_available() execute on an internal Routing
+     * Service thread. The custom dispose() operation doesn't run on that
+     * thread. Since the take() and return_loan() operations also need to access
+     * the data_samples_ list this protection is required.
+     */
+    std::lock_guard<std::mutex> guard(data_samples_mutex_);
+    std::copy(
+            data_samples_.begin(),
```
alexcamposruiz: See my comment in the constructor. If you make data_samples_ a vector of unique_ptr, you may need to change this to:

 ```
std::transform(
                    data_samples_.begin(),
                    data_samples_.end(),
                    std::back_inserter(stream),
                    [](const auto& element) { return element.get(); }
```

Comment (alexcamposruiz) [reply examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx]: That obtains the underlying pointer from the unique_ptr for each element

File: examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.hpp [hunk 60]
```diff
@@ -0,0 +1,76 @@
...
+     * CSV formatted file one line at a time. The file only contains data and 
+     * no meta data information.
+     */
+    void file_reading_thread();
+
+    FileConnection *file_connection_;
+    rti::routing::adapter::StreamReaderListener *reader_listener_;
+    std::thread filereader_thread_;
+    bool stop_thread_;
+    uint16_t sampling_period_;
```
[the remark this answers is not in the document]
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,152 @@
...
+    std::getline(s, shapesize, ',');
+
+    /**
+     * Note that we read one line at a time from the CSV file in the
+     * function file_reading_thread()
+     */
+    samples.resize(1);
+    infos.resize(1);
+
+    DynamicData *sample = new DynamicData(*adapter_type_);
```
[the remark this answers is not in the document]
pralhadsapre: Good point. Done.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,152 @@
...
+    return;
+}
+
+void FileStreamReader::take(
+        std::vector<dds::core::xtypes::DynamicData *> &samples,
+        std::vector<dds::sub::SampleInfo *> &infos,
+        const SelectorState &selector_state)
+{
+    take(samples, infos);
+    return;
```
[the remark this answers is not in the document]
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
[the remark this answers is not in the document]
pralhadsapre: I ended up creating a unique_ptr and passing new in its constructor.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,152 @@
...
+        delete infos[i];
+    }
+    samples.clear();
+    samples.clear();
+}
+
+FileStreamReader::~FileStreamReader()
+{
+    stop_thread_ = true;
+    filereader_thread_.join();
```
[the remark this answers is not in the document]
pralhadsapre: Done. I added a helper function called shutdown_file_reader_thread().

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,152 @@
...
+
+    filereader_thread_ =
+            std::thread(&FileStreamReader::file_reading_thread, this);
+}
+
+void FileStreamReader::take(
+        std::vector<dds::core::xtypes::DynamicData *> &samples,
+        std::vector<dds::sub::SampleInfo *> &infos)
+{
+    std::istringstream s(buffer_);
```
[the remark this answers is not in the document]
pralhadsapre: Yes. I added a mutex for that.

examples/routing_service/routing_service_file_adapter/cpp/FileStreamReader.cxx
```diff
@@ -0,0 +1,152 @@
...
+    std::string color;
+    std::string x;
+    std::string y;
+    std::string shapesize;
+
+    // Reading input from the CSV file
+    std::getline(s, color, ',');
```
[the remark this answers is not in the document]
pralhadsapre: Done.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
pralhadsapre: Done. We can't directly call clear(). I am using erase() with a range specified.

examples/routing_service/routing_service_file_adapter/cpp/FileInputDiscoveryStreamReader.cxx
pralhadsapre: Done.

Issue #174, opened 2019-01-07
Port the examples to Cake


iblancasa:
This issue is similar to #161 but for C#. [Cake](https://cakebuild.net/) is a build system for C#.

PR #175, opened 2019-01-07
[WIP] Port the C# examples to Cake

iblancasa:
This PR solves #174.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,13 @@ makefile*
 *.sln
 *.vcxproj
 
+# Cake projects
+tools/*dll
+*exe
+*md5sum
+src
+Cake/
+
 # IDE tools files
 .clang_complete
 .linter-clang-flags
--- a/resources/cake/ConnextDDSExamples.cake
+++ b/resources/cake/ConnextDDSExamples.cake
@@ -1,0 +1,134 @@
+/*******************************************************************************
+* (c) 2019 Copyright, Real-Time Innovations, Inc.  All rights reserved.
+* RTI grants Licensee a license to use, modify, compile, and create derivative
+* works of the Software.  Licensee has the right to distribute object form only
+* for use with RTI products.  The Software is provided "as is", with no
+* warranty of any type, including any warranty for fitness for any purpose. RTI
+* is under no obligation to maintain or support the Software.  RTI shall not be
+* liable for any incidental or consequential damages arising out of the use or
+* inability to use the software.
+******************************************************************************/
+
+var target = Argument ("target", "Default");
+var rtiArchitecture = Argument ("ConnextDDSArchitecture", "");
+var configuration = Argument("Configuration", "Release");
+
+/*
+ * Task to run Codegen and generate the the source code files and the
+ * Visual Studio Solution.
+ *
+ * This tasks expects the variables:
+ *     - NDDSHOME environment variable: set to the NDDSHOME folder
+ *     - idlName: path to the IDL file to process.
+ *     - rtiArchitecture: ConnextDDS architecture to build.
+ *
+ * All the generated files will be placed under the "src" folder. The "src"
+ * folder will be created under the current working directory.
+ */
+Task("Codegen")
+    .Does(() => {
+        string nddshomeValue = Environment.GetEnvironmentVariable("NDDSHOME");
+
+        if (string.IsNullOrEmpty(nddshomeValue)) {
+            throw new Exception("NDDSHOME environment variable was not set");
+        } else if (string.IsNullOrEmpty(rtiArchitecture)) {
+            throw new Exception("The RTI Architecture must be provided");
+        }
+
+        DirectoryPath nddshome = new DirectoryPath(nddshomeValue);
+        string rtiddsgen = "rtiddsgen";
+
+        if (IsRunningOnWindows()) {
+            rtiddsgen += ".bat";
+        }
+
+        string codegen = nddshome.Combine("bin").Combine(rtiddsgen).ToString();
+        string lang = "-language C#";
+        DirectoryPath cwd = Environment.CurrentDirectory;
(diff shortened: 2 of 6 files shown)


resources/cake/ConnextDDSExamples.cake
```diff
@@ -0,0 +1,134 @@
...
+* RTI grants Licensee a license to use, modify, compile, and create derivative
+* works of the Software.  Licensee has the right to distribute object form only
+* for use with RTI products.  The Software is provided "as is", with no
+* warranty of any type, including any warranty for fitness for any purpose. RTI
+* is under no obligation to maintain or support the Software.  RTI shall not be
+* liable for any incidental or consequential damages arising out of the use or
+* inability to use the software.
+******************************************************************************/
+
+var target = Argument ("target", "Default");
```
pleonex: the other two arguments starts with upper char, I think we should follow the same convention or make all of the lower char

resources/cake/ConnextDDSExamples.cake
```diff
@@ -0,0 +1,134 @@
...
+var rtiArchitecture = Argument ("ConnextDDSArchitecture", "");
+var configuration = Argument("Configuration", "Release");
+
+/*
```
pleonex: In C# the `/* */` style comments are not usually used. To comment a method the style is to use comments starting with `///` and with an XML format like:
```cs
/// <summary>
/// Something.
/// </summary>
```
For regular comments or even the header comments, it's usually use the regular line comment `//`. The Cake Tasks I think are similar to methods so I would use the `///` style comments.

resources/cake/ConnextDDSExamples.cake
```diff
@@ -0,0 +1,134 @@
...
+* works of the Software.  Licensee has the right to distribute object form only
+* for use with RTI products.  The Software is provided "as is", with no
+* warranty of any type, including any warranty for fitness for any purpose. RTI
+* is under no obligation to maintain or support the Software.  RTI shall not be
+* liable for any incidental or consequential damages arising out of the use or
+* inability to use the software.
+******************************************************************************/
+
+var target = Argument ("target", "Default");
+var rtiArchitecture = Argument ("ConnextDDSArchitecture", "");
```
pleonex: If you use the overload without the second argument (the default value), Cake will automatically throws an exception if it's missing.

resources/cake/ConnextDDSExamples.cake
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        DirectoryPath nddshome = new DirectoryPath(nddshomeValue);
+        string rtiddsgen = "rtiddsgen";
+
+        if (IsRunningOnWindows()) {
+            rtiddsgen += ".bat";
+        }
+
+        string codegen = nddshome.Combine("bin").Combine(rtiddsgen).ToString();
```
pleonex: If I am not wrong Cake implements an implicit casting from `string` to `DirectoryPath` so I think you can simplify these path handling to:
```cs
string codegen = $"{nddshomeValue}/bin/{rtiddsgen}"
```

The same for other variables like `outputDir`, `idl`, ...
Also Cake will convert the directory path for Windows automatically like CMake does.

resources/cake/ConnextDDSExamples.cake
```diff
@@ -0,0 +1,134 @@
...
+        } else {
+            CleanDirectory(outputDir);
+        }
+        string outputArg = "-d " + outputDir;
+
+        string idlName = GetFiles("*idl").First().GetFilenameWithoutExtension().ToString();
+
+        string idl = cwd.Combine(idlName + ".idl").ToString();
+
+        string args = String.Format("{0} {1} -ppDisable {2} {3}",
```
pleonex: There is a new string formatting that I think it's easier to read:
```csharp
string args = $"{lang} {example} -ppDisable {idl} {outputArg}";
```

resources/cake/ConnextDDSExamples.cake
```diff
@@ -0,0 +1,134 @@
...
+                                    lang, example, idl, outputArg);
+
+        ProcessSettings settings = new ProcessSettings { Arguments = args };
+
+        using(var process = StartAndReturnProcess(codegen, settings)) {
```
pleonex: I think you can simplify this to:
```csharp
    var retcode = StartProcess(codegen, args);
    if (retcode != 0) {
        throw new Exception("Fail to run rtiddsgen.");
    }
```

resources/cake/ConnextDDSExamples.cake
```diff
@@ -0,0 +1,134 @@
...
+        var found = GetFiles(pattern);
+
+        if (found.Count() == 0) {
+            throw new Exception("No SLN files were found");   
+        }
+
+        var sln = found.First();
+        string slnName = sln.GetFilename().ToString();
+        
+        // MSBuild(slnName, settings =>
```
pleonex: I think you forgot to remove / uncomment these lines

resources/cake/ConnextDDSExamples.cake
```diff
@@ -0,0 +1,134 @@
...
+    });
+
+/*
+ * Add the example.
+ */
+Task("AddConnextDDSExample")
+    .IsDependentOn("Codegen")
+    .IsDependentOn("Copy")
+    .IsDependentOn("Build")
+    .Does(() => {
```
pleonex: You can remove the `Does()`, a task can just have `IsDependtOn()`

.gitignore
```diff
@@ -38,6 +38,13 @@ makefile*
 *.sln
 *.vcxproj
 
+# Cake projects
+tools/*dll
```
pleonex: [Cake recommends](https://cakebuild.net/docs/tutorials/pinning-cake-version#gitignore) the following exclusion patterns:
```
tools/*
!tools/packages.config
```

PR #176, opened 2019-01-17
Make the example more CMake friendly

iblancasa:
I tried to do some small improvements.

@titosanxez, please, could you take a look? Thank you!

### DIFF
--- a/examples/routing_service_shapes_processor/cpp/CMakeLists.txt
+++ b/examples/routing_service_shapes_processor/cpp/CMakeLists.txt
@@ -8,15 +8,19 @@
 #  RTI is under no obligation to maintain or support the Software.  RTI shall
 #  not be liable for any incidental or consequential damages arising out of the
 #  use or inability to use the software.
-#
 
 cmake_minimum_required(VERSION 3.11)
-set (CMAKE_CXX_STANDARD 11)
-project (ShapesProcessor)
+project(ShapesProcessor)
 
+if(NOT BUILD_SHARED_LIBS)
+    set(msg
+        "Static linking is not supported for this example. You can build it "
+        "using the -DBUILD_SHARED_LIBS=1 CMake parameter")
+    message(FATAL_ERROR ${msg})
+endif()
 
 # Find RTI Connext dependencies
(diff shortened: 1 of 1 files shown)


examples/routing_service_shapes_processor/cpp/CMakeLists.txt
```diff
@@ -8,15 +8,19 @@
...
 #  not be liable for any incidental or consequential damages arising out of the
 #  use or inability to use the software.
-#
 
 cmake_minimum_required(VERSION 3.11)
-set (CMAKE_CXX_STANDARD 11)
-project (ShapesProcessor)
+project(ShapesProcessor)
 
+if(NOT BUILD_SHARED_LIBS)
```
titosanxez: Nice. Thanks for this

examples/routing_service_shapes_processor/cpp/CMakeLists.txt
```diff
@@ -8,15 +8,19 @@
...
+    set(msg
+        "Static linking is not supported for this example. You can build it "
+        "using the -DBUILD_SHARED_LIBS=1 CMake parameter")
+    message(FATAL_ERROR ${msg})
+endif()
 
 # Find RTI Connext dependencies
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CONNEXTDDS_DIR}/resource/cmake")
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "$ENV{NDDSHOME}/resource/cmake")
```
titosanxez: why this? aren't we using CONNEXT_DIR in findConnextDDs? 
Also we would need to update the example isnructions to reflect this change

iblancasa: Actually I can revert this. I set the NDDSHOME environment variable because usually is what the people sets following the instructions. But I don't have a strong opinion about. So, what do you prefer?

titosanxez: I don't have a strong opinion either though I'm not a fun of environment variables. Ideally it's nice if we can live without NDDSHOME (also the name is somewhat old). And since we already define CONNEXT_DIR, we could embrace it.

It also depends on what we want to do with other examples as well.

Issue #161, opened 2018-12-29
Port the examples to CMake


iblancasa:
It would be amazing to port the examples to CMake. That would make easier to build the examples.

PR #177, opened 2019-01-24
Porting Recording Service examples from main repository to RTI community examples

juanlu-rti:
The port includes changing from makefiles and manually-maintained VS projects to CMake.

Internal process: GALLEON-179.


examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+add_library(
+    FileStorageWriter MODULE
+    ${CMAKE_SOURCE_DIR}/FileStorageWriter.c
+    ${CMAKE_SOURCE_DIR}/FileStorageWriter.h
+)
+
+target_link_libraries(
+    FileStorageWriter
+    RTIConnextDDS::routing_service_infrastructure
+    RTIConnextDDS::c_api
```
iblancasa: The ``c_api`` library is also linked when you link against the ``RTIConnextDDS::routing_service_infrastructure`` target

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+    ${CONNEXTDDS_EXTERNAL_LIBS}
+)
+
+target_include_directories(
+    FileStorageWriter
+    PUBLIC ${CMAKE_BINARY_DIR}
```
iblancasa: The ``PUBLIC`` visibility is used when you are linking against the target (in this case, the ``FileStorageWriter`` library). 

You created the ``FileStorageWriter`` library as a ``MODULE`` library.

From the CMake documentation:
```
MODULE libraries are plugins that are not linked into other targets but may be loaded dynamically at runtime using dlopen-like functionality.
```

So, this means nobody will link against the ``FileStorageWriter`` library. For that reason, you should use ``PRIVATE`` visibility.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+    PUBLIC ${CONNEXTDDS_INCLUDE_DIRS}
```
iblancasa: This one is added automatically when you link against the ``RTIConnext::`` targets.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+
+target_include_directories(
+    HelloMsg_publisher
+    PUBLIC ${CMAKE_BINARY_DIR}
+    PUBLIC ${CONNEXTDDS_INCLUDE_DIRS}
+)
+
+target_compile_definitions(
+    HelloMsg_publisher
+    PUBLIC ${CONNEXTDDS_COMPILE_DEFINITIONS}
```
iblancasa: This is automatically added  when you link against the ``RTIConnext::`` targets

examples/recording_service/pluggable_storage/c/CMakeLists.txt
titosanxez: That's not necessarily true.  The rsinfrastructure is currently a standalone library. Unless the example is ussing DDS/core stuff, is not really needed.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+)
+
+# Generate type plug-in files for the storage type
+add_custom_command(
+    OUTPUT 
+        ${CMAKE_SOURCE_DIR}/HelloMsg.c 
+        ${CMAKE_SOURCE_DIR}/HelloMsgPlugin.c
+        ${CMAKE_SOURCE_DIR}/HelloMsgSupport.c
+    MAIN_DEPENDENCY ${CMAKE_SOURCE_DIR}/HelloMsg.idl
+    COMMAND ${RTICODEGEN} -language C -replace -d ${CMAKE_SOURCE_DIR} ${CMAKE_SOURCE_DIR}/HelloMsg.idl
```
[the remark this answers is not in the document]
titosanxez: It's the first time I see the MAIN_DEPENDENCY in the examples. Iis this they way to call rtiddsgen or do we have some other way/convention? @iblancasa

examples/recording_service/pluggable_storage/c/FileStorageReader.h
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Entry-point function to the C reader plugin. This function will create an
+ * instance of the StorageReader example. This example opens a file passed in by
+ * the configuration in the storage property tag. The file contains samples in
+ * a textual format that is understood by the example StorageReader classes.
+ * This example works exclusively with the type defined in the provided IDL
+ * file, HelloMsg.idl.
+ */
+DLLExport
```
titosanxez: we should use the RTI_USER_DLL_EXPORT

examples/recording_service/pluggable_storage/c/HelloMsg_subscriber.c
```diff
@@ -0,0 +1,305 @@
+
```
titosanxez: This file is exactly as the generated one, isn;t it? if so we should remove it

examples/recording_service/pluggable_storage/cpp/FileStorageReader.hpp
```diff
@@ -0,0 +1,178 @@
...
+     * data stream has been read and that we're ready for termination.
+     */
+    virtual bool finished();
+
+    virtual void reset();
+
+private:
+
+    std::ifstream *data_file_;
+
```
titosanxez: we should remove the blank liknes between attributes

examples/recording_service/pluggable_storage/cpp/FileStorageReader.hpp
```diff
@@ -0,0 +1,178 @@
...
+ * Note: this example is only fit to work with the provided type definition
+ * (type HelloMsg - see HelloMsg.idl for more information).
+ */
+class FileStorageStreamReader :
+        public rti::recording::storage::DynamicDataStorageStreamReader {
+public:
+
+    FileStorageStreamReader(std::ifstream *data_file);
+
+    virtual ~FileStorageStreamReader();
```
titosanxez: we don't need the virtual keyword in this class, since there is no further inheritance

examples/recording_service/pluggable_storage/cpp/FileStorageWriter.cxx
```diff
@@ -0,0 +1,205 @@
...
+        }
+        /* Time was returned in seconds. Transform to nanoseconds */
+        current_time *= NANOSECS_PER_SEC;
+        info_file_ << "End timestamp: " << current_time << std::endl;
+    }
+}
+
+rti::recording::storage::StorageStreamWriter *
+FileStorageWriter::create_stream_writer(
+            const rti::routing::StreamInfo& stream_info,
```
titosanxez: indentation

examples/recording_service/pluggable_storage/c/CMakeLists.txt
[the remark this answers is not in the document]
iblancasa: We were not using it because it has some side effects. It is almost like "DEPENDS" but will do more things for VS.


Actually, something to improve would be:
* Create a custom target depending on the output files generated by the custom command  
* The libraries and the applications should depends on that custom

That will ensure Codegen is not called twice in multi-job builds (if Codegen is called twice, the build will break)

examples/recording_service/pluggable_storage/c/CMakeLists.txt
iblancasa: Ok, so it is something we need to improve in the "FindConnextDDS" CMake script. Right now, it is included always.

iblancasa:
This example needs to be built with C++11, right?

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+
+find_package(
+    RTICodeGenerator "3.0.0"
+    REQUIRED
+)
+
+# Generate type plug-in files for the storage type
+add_custom_command(
+    OUTPUT 
+        ${CMAKE_SOURCE_DIR}/HelloMsg.c
```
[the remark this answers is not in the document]
juanlu-rti: True. Will add them.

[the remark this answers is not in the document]
juanlu-rti: @titosanxez , I took a look at a Routing Service example for inspiration on some things, you may want to also review this issue in them?

examples/recording_service/pluggable_storage/c/CMakeLists.txt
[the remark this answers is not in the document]
juanlu-rti: Okay, so I'll create the custom target approach. What about the MAIN_DEPENDENCY, should I change it to DEPENDS?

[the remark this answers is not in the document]
iblancasa: I don't have a strong opinion regarding  the ``MAIN_DEPENDENCY`` option.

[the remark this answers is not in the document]
jpovedano: I don't have an strong opinion either, but it looks, that when there is a 1-1 dependency relationship (e.g. file generation), MAIN_DEPENDENCY might be better:

https://rix0r.nl/blog/2015/08/13/cmake-guide/

examples/recording_service/pluggable_storage/c/CMakeLists.txt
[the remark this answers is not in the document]
juanlu-rti: My last commit adds quotes as per @iblancasa 's comment.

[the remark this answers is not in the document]
juanlu-rti: OK, will change references.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
juanlu-rti: I have not understood this correctly. This is actually a library that will not be linked into other targets but will effectively be dl-open-loaded by Recorder (the same as FileStorageReader will be dl-open-loaded by Replay and Converter), so I do not understand why MODULE doesn't fit?

examples/recording_service/pluggable_storage/c/CMakeLists.txt
juanlu-rti: Oh, I see, so I can remove the {{CONNEXTDDS_INCLUDE_DIRS}} from this CMake command, right? I'll try it out and push.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
juanlu-rti: I have committed changes that remove the unnecessary CONNEXTDDS_COMPILE_DEFINITIONS and CONNEXTDDS_INCLUDE_DIRS from the CMakeLists.txt files (both C and C++).

examples/recording_service/pluggable_storage/c/HelloMsg_subscriber.c
juanlu-rti: Hmmmm, I know the publisher varies, but it's probably true that this is just the same as the generated one. I'll remove it.

examples/recording_service/pluggable_storage/c/FileStorageReader.h
juanlu-rti: Done.

examples/recording_service/pluggable_storage/c/HelloMsg_subscriber.c
juanlu-rti: @titosanxez , I realised that in both subscribers we do modify the topic name. So I believe we cannot generate over them and we cannot remove them from git. That should be fine, though.

examples/recording_service/pluggable_storage/cpp/FileStorageReader.hpp
juanlu-rti: I will. On it.

examples/recording_service/pluggable_storage/cpp/FileStorageReader.hpp
juanlu-rti: That's true. Will remove.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,140 @@
...
+#  only for use with RTI products.  The Software is provided "as is", with no
+#  warranty of any type, including any warranty for fitness for any purpose.
+#  RTI is under no obligation to maintain or support the Software.  RTI shall
+#  not be liable for any incidental or consequential damages arising out of the
+#  use or inability to use the software.
+#
+cmake_minimum_required(VERSION 3.11)
+
+set(CMAKE_SYSTEM_NAME "Windows")
+set(CMAKE_SYSTEM_PROCESSOR x86)
```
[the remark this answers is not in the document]
juanlu-rti: It was removed.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,140 @@
...
+
+project (PluggableStorage)
+set(CONNEXTDDS_DISABLE_VERSION_CHECK TRUE)
```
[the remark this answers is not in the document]
juanlu-rti: It was removed.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,135 @@
...
+)
+
+set(GENERATED_CODE
+    "${CMAKE_CURRENT_SOURCE_DIR}/HelloMsg.c"
+    "${CMAKE_CURRENT_SOURCE_DIR}/HelloMsgPlugin.c"
+    "${CMAKE_CURRENT_SOURCE_DIR}/HelloMsgSupport.c"
+)
+
+# Generate type plug-in files for the storage type
+add_custom_command(
```
titosanxez: @iblancasa , should we be using the new way to generate code? or isn't available for Galleon?

juanlu-rti:
@titosanxez , @iblancasa , I believe the code is quite ready to be merged. Let me know if you have any more questions or change requests. Otherwise, feel free to merge :)

examples/recording_service/pluggable_storage/c/CMakeLists.txt
iblancasa: It will be available in this repository as soon as we merge #161. The idea is to refactor the examples after that merge to use the new way to generate code.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,135 @@
...
+add_custom_target(
+    generated_code DEPENDS ${GENERATED_CODE}
+)
+
+# Define the library that will provide the storage writer plugin
+add_library(
+    FileStorageWriter MODULE
+        "${CMAKE_CURRENT_SOURCE_DIR}/HelloMsg.c" 
+        "${CMAKE_CURRENT_SOURCE_DIR}/FileStorageWriter.c"
+        "${CMAKE_CURRENT_SOURCE_DIR}/FileStorageWriter.h"
```
iblancasa: This is a header file. This should be included using the ``target_include_directories`` CMake buil-in macro.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,135 @@
...
+    FileStorageWriter
+    RTIConnextDDS::routing_service_infrastructure
+    RTIConnextDDS::c_api
+    ${CONNEXTDDS_EXTERNAL_LIBS}
+)
+
+target_include_directories(
+    FileStorageWriter
+    PUBLIC "${CMAKE_BINARY_DIR}"
+    PUBLIC ${CONNEXTDDS_INCLUDE_DIRS}/recordingservice
```
iblancasa: You can add a list of directories after the visibility word. I mean:
```cmake
PUBLIC
    dir1
    dir2
```

iblancasa: Add quotes to this path

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,135 @@
...
+target_link_libraries(
+    FileStorageWriter
+    RTIConnextDDS::routing_service_infrastructure
+    RTIConnextDDS::c_api
+    ${CONNEXTDDS_EXTERNAL_LIBS}
+)
+
+target_include_directories(
+    FileStorageWriter
+    PUBLIC "${CMAKE_BINARY_DIR}"
```
iblancasa: Use ``CMAKE_CURRENT_BINARY_DIR`` instead. If somebody includes this project as sub-project, the build will fail.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,135 @@
...
+    "${CMAKE_CURRENT_SOURCE_DIR}/HelloMsgPlugin.c"
+    "${CMAKE_CURRENT_SOURCE_DIR}/HelloMsgSupport.c"
+)
+
+# Generate type plug-in files for the storage type
+add_custom_command(
+    OUTPUT ${GENERATED_CODE}
+    MAIN_DEPENDENCY "${CMAKE_CURRENT_SOURCE_DIR}/HelloMsg.idl"
+    COMMAND ${RTICODEGEN} -language C -update typefiles
+            -platform ${CONNEXTDDS_ARCH} -d "${CMAKE_CURRENT_SOURCE_DIR}"
```
iblancasa: Instead of modify the source directory, it is a better idea to store the generated code in the binary dir. Probably (to make everything easier to undertand), in a sub-directory of the binary directory.

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -0,0 +1,135 @@
...
+        core
+        routing_service
+)
+
+find_package(
+    RTICodeGenerator "3.0.0"
+    REQUIRED
+)
+
+set(GENERATED_CODE
```
iblancasa: Probably, it would be a good idea to set a more significant name to this variable

juanlu-rti:
Feedback applied. @iblancasa , I have applied the same feedback to the C++ example. 

I believe anybody with write permission can go ahead and merge.

examples/recording_service/service_admin/cpp/README.rst
```diff
@@ -0,0 +1,143 @@
...
+The following files are part of the example:
+
+- ``Requester.hpp`` and ``Requester.cxx``: these files contain the example C++
+  code that prepares and sends the command request and then waits for a reply
+  from the running service instance.
+
+- ``USER_QOS_PROFILES.xml``: this file defines DDS XML QoS profiles to be used
+  by the application and by *Recording Service*.
+
+- ``recorder_remote_admin.xml``: this is a configuration file for Recording
```
titosanxez: We need a file name convention for configuration files @iblancasa @fgaranda

examples/recording_service/service_admin/cpp/README.rst
```diff
@@ -0,0 +1,143 @@
...
+===================
+
+First of all, we need a running |RS| instance with remote administration enabled.
+We will use the |RS| configuration shipped with the example, called 
+``recorder_remote_admin.xml``:
+
+.. code-block:: text
+
+    cd <binary directory>
+    $(CONNEXTDDS_DIR)/bin/rtirecordingservice -cfgFile recorder_remote_admin.xml -cfgName remote_admin
```
titosanxez: You can use the multiline format with '\', and indent twice the parameters in new line

examples/recording_service/service_admin/cpp/README.rst
```diff
@@ -0,0 +1,143 @@
...
+Once the |RS| instance is running, it is ready to receive remote command requests.
+The administration app receives the following command-line parameters:
+
+- The kind of command to be sent, following a *CRUD* pattern (CREATE, READ, 
+  UPDATE, DELETE).
+- The identifier of the resource we're accessing, 
+  e.g. ``recording_services/remote_admin/state``
+- [Optional] The arguments that the command kind needs for performing the action,
+  e.g. ``paused`` (if the objective state of the service is *paused*)
+- [Optional] ``--domain-id``: this parameter should be used when the |RS|
```
titosanxez: In general we don't use the '--param-name' format but '-paramName' instead. It may be ok for this example but we should aim for consistency.

examples/recording_service/service_admin/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+
+find_package(
+    RTICodeGenerator "3.0.0"
+    REQUIRED
+)
+
+# Create the directory that's going to contain the generated code files
+add_custom_target(
+    generated_code_dir ALL
+    COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_BINARY_DIR}/generated"
```
iblancasa: Use ``CMAKE_CURRENT_BINARY_DIR`` instead of ``CMAKE_BINARY_DIR``.

examples/recording_service/service_admin/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+
+target_include_directories(
+    Requester
+    PUBLIC "${CMAKE_BINARY_DIR}/generated"
+)
+
+target_link_libraries(
+    Requester
+    RTIConnextDDS::cpp2_api
+    RTIConnextDDS::messaging_cpp2_api
```
iblancasa: The ``RTIConnextDDS::messaging_cpp2_api`` imported target includes the ``RTIConnextDDS::cpp2_api`` imported target. So, you don't need to link against the ``RTIConnextDDS::cpp2_api`` imported target.

juanlu-rti:
@titosanxez about the command-line parameters naming, I certainly see your point. @fgaranda and I discussed it and we liked this approach (the one I finally used), but we definitely understood that this yields a less consistent example.
If you guys want, I can change the parameter names, no big deal.

examples/recording_service/service_admin/cpp/README.rst
titosanxez: @juanlu-rti @fgaranda  yes I can see why you guys like it better ( I do too) but at this point we need to make sure we follow a pattern, which should be also the same for examples. Right now we are probably not consistent in the examples and we should start working towards that. I would make the change if you have a chance if we agree this is the format we should be doing (which is the format of the infsvcs for example).

PR #179, opened 2019-02-04
Incorporate feedback from hands-on

titosanxez:


examples/routing_service_shapes_processor/cpp/README.rst
```diff
@@ -21,6 +21,7 @@ From the directory containing the example sources:
     cd build
     cmake -DCONNEXTDDS_DIR=<Connext directory> \
             -DCONNEXTDDS_ARCH=<ARCH> \
+            -DCMAKE_RUNTIME_OUTPUT_DIRECTORY
```
iblancasa: This needs a value and is something you  can set as a target property

titosanxez: Good catch, I forgot to update this properly. Will this be enough for Visual Studio?

examples/routing_service_shapes_processor/cpp/CMakeLists.txt
```diff
@@ -33,16 +44,27 @@ find_package(
 add_library(shapesprocessor
     "${CMAKE_CURRENT_SOURCE_DIR}/ShapesProcessor.cxx")
 
+# Set target properties for lang requirement output library name
+set_target_properties(shapesprocessor
+    PROPERTIES
+        CXX_STANDARD 11
+        OUTPUT_NAME_DEBUG shapesprocessord
+        LIBRARY_OUTPUT_DIRECTORY "${output_dir}"
```
iblancasa: Do you need to specify all the configurations? I mean, if you set ``LIBRARY_OUTPUT_DIRECTORY`` you shouldn't need to set ``LIBRARY_OUTPUT_DIRECTORY_RELEASE`` . If ``LIBRARY_OUTPUT_DIRECTORY_RELEASE`` is not set, CMake will try to use ``LIBRARY_OUTPUT_DIRECTORY``.

titosanxez: We discussed this online and we agreed we need them

PR #183, opened 2019-03-15
Fixing several issues with examples for 6.0.1 (GMAINT-135 GMAINT-134 GMAINT-124)

juanlu-rti:
@iblancasa , could you review these changes? This was detected by Ismael when building in VS. Before the change it was building with the same library for both modes (release or debug). So there was a correct build given the specified mode, but you could not have both correct builds (debug/release) at the same time.


juanlu-rti:
@titosanxez , I have fixed GMAINT-135, GMAINT-134 and GMAINT-124. For GMAINT-135 you should see RECORD-1113, as I found a related product issue.

PR #195, opened 2019-04-29
Added python script to upload rti_connextdds code to amazon bucket.

lulivi:
Also deleted `.gitignore`s from two examples because there is already a `build` folder ignore in root `.gitignore`.


resources/aws/rti_aws_uploader.py
```diff
@@ -0,0 +1,264 @@
...
+    try:
+        s3.meta.client.head_bucket(Bucket=bucket_name)
+    except botocore.client.ClientError as e:
+        if e.response['Error']['Code'] == '404':
+            exists = False
+            error = f'Bucket {bucket_name} not found.'
+    except botocore.exceptions.NoCredentialsError:
+        exists = False
+        error = 'Credentials not found.'
+    return exists, error
```
iblancasa: Returning an error message to handle it is not a good practice. It would be better to create your own `Exception` class and raise an object of this new class.

resources/aws/rti_aws_uploader.py
```diff
@@ -0,0 +1,264 @@
...
+
+    if args.cfg_path:
+        cfg_path = args.cfg_path.resolve()
+        if not cfg_path.is_file():
+            sys.exit(f'Error: Path "{cfg_path}" does not exist.')
+        os.environ['AWS_SHARED_CREDENTIALS_FILE'] = str(cfg_path)
+
+    s3 = boto3.resource('s3')
+
+    exists, error = bucket_exists(s3, args.bucket_name)
```
iblancasa: You can return directly the `exists` value and do something like:
```python
if not bucket_exists(s3, args.bucket_name):
```

resources/aws/rti_aws_uploader.py
```diff
@@ -0,0 +1,264 @@
...
+import sys
+import boto3
+import botocore
+import argparse
+
+from pathlib import Path
+from typing import List, Set
+
+
+def bucket_exists(s3, bucket_name: str):
```
iblancasa: Type hinting for the return value

iblancasa: Type hinting for `s3`

resources/aws/rti_aws_uploader.py
```diff
@@ -0,0 +1,264 @@
...
+    """Checks if bucket exists.
+
+    Args:
+        s3 (boto3.resource('s3')): s3 resource to check the bucket from.
+        bucket_name (str): Selected bucket name to check.
+
+    Returns:
+        exists (bool): True if the bucket exists.
```
iblancasa: Since you are using type hinting for all the function, you don't need to specify the type here. Sphinx and other processors are smart enough to get the type.

resources/aws/rti_aws_uploader.py
```diff
@@ -0,0 +1,264 @@
+#!/usr/bin/env python3.7
```
iblancasa: It would be nice if you could use the [`logging` module](https://docs.python.org/3/library/logging.html). You can add a `debug` mode to make easier fixing issues.

resources/aws/rti_aws_uploader.py
```diff
@@ -0,0 +1,264 @@
...
+        if child.is_dir():
+            file_list.extend(
+                _get_directory_tree_file_path_list(child.resolve(),
+                                                   filtered_out_subdirs))
+        else:
+            file_list.append(child.resolve())
+    return sorted(file_list)
+
+
+def upload_directory(s3,
```
iblancasa: I think you followed a different format for the parameters in other methods.

resources/aws/rti_aws_uploader.py
```diff
@@ -0,0 +1,264 @@
...
+        selected_top_items (Set[Path]): selected top items list (default=None).
+        filtered_out_subdirs (Set[Path]): filtered out subdir list
+            (default=None).
+    """
+    bucket = s3.Bucket(bucket_name)
+    file_path_filtered_list = _get_filtered_file_list(
+        local_dir_path, selected_top_items, filtered_out_subdirs)
+    print('The following files will be uploaded:')
+    print(*file_path_filtered_list, sep='\n')
+    if input('Are you sure? (yes/no): ').strip() != 'yes':
```
iblancasa: This should not go here. Why? If I am using the `upload_directory` function, it will be totally unexpected to be blocked by an user interaction.

resources/aws/rti_aws_uploader.py
```diff
@@ -0,0 +1,264 @@
...
+    else:
+        print(f'Error: Folder "{remote_dir_name}" not found.')
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description=__doc__,
+        allow_abbrev=False,
+        formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument('-b',  type=str,
```
iblancasa: I think it would be better to use descriptive names for the parameters.

iblancasa:
Good job!

resources/aws/rti_aws_uploader.py
lulivi: Please @iblancasa take a look to the new changes.

PR #202, opened 2019-05-21
Git hooks and continuous integration

lulivi:
- [x] Add git hooks for local linting.
- [x] Add travis integration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: Xenial
-language: node_js
-node_js:
-  - "6"
+language: python
+python: 
+  - "3.6"
 before_install:
   - wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
   - sudo add-apt-repository -y 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial main'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: Xenial
 language: python
 python: 
-  - "3.6"
+  - "3.7"
 before_install:
   - wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
   - sudo add-apt-repository -y 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial main'
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ before_install:
   - wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
   - sudo add-apt-repository -y 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial main'
   - sudo apt update
-  - sudo apt-get install -y clang-format-9
+install:
+  - sudo apt install -y clang-format-9
   - npm install -g markdownlint-cli
 script:
   - ./githooks/pre-commit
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -0,0 +1,12 @@
+dist: Xenial
+language: node_js
+node_js:
+  - "6"
+before_install:
+  - wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
+  - sudo add-apt-repository -y 'deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial main'
+  - sudo apt update
+  - sudo apt-get install -y clang-format-9
```
iblancasa: I think you should run ` sudo apt-get install -y clang-format-9` and `npm install -g markdownlint-cli` during the `install` stage.

CONTRIBUTING.md
```diff
@@ -1,12 +1,18 @@
...
-In order to accept your pull request, we need you to sign a Contributor License Agreement (CLA). Complete your CLA here: http://community.rti.com/cla. You only need to do this once, we cross-check your Github username with the list of contributors who have signed the CLA.
-
-### License
-RTI grants Licensee a license to use, modify, compile, and create derivative works 
-of the Software.  Licensee has the right to distribute object form only for use with RTI 
-products.  The Software is provided “as is”, with no warranty of any type, including 
-any warranty for fitness for any purpose. RTI is under no obligation to maintain or 
-support the Software.  RTI shall not be liable for any incidental or consequential 
-damages arising out of the use or inability to use the software.
+## Contributor License Agreement (CLA)
```
iblancasa: @fgaranda it would be really nice to finally integrate https://www.clahub.com/ in the repository

githooks/README.md
```diff
@@ -0,0 +1,30 @@
...
+Checkout [llvm build page](http://llvm.org/builds/) for info about downloading
+`clang-format`. Version 9 of clang-format must be installed along with it's git
+integration.
+
+## markdownlint
+
+To install last version of `markdownlint`:
+
+```bash
+npm install -g markdownlint-cli --save-dev
```
iblancasa: You don't need to add `--save-dev` since you are not creating a NodeJS application where `markdownlint-cli` is a dependency. Even, it shouldn't be needed to use `-g` (it will be installed in a local folder called `node_modules` but that path will not be added to the `PATH` environment variable).

iblancasa: Document that NodeJS is needed and the minimum required version.

githooks/README.md
```diff
@@ -0,0 +1,30 @@
+# Git hook for local pre-commit linting
```
iblancasa: Can this be executed on Windows?

lulivi: It should, with the correct installation of `markdownlint` and `git-clang-format`.

githooks/README.md
lulivi: How can I execute any package installing it without `-g` option? I tried `./node_modules/.bin/<package>` but didn't work.

iblancasa: Oh! You should be able. But that is not the important part. What is important is to remove `--save-dev`

githooks/README.md
iblancasa: Perfect!

githooks/README.md
lulivi: All right!

CONTRIBUTING.md
fgaranda: Sounds good.

PR #204, opened 2019-05-22
Update example/recording service recording partitions in groups

guyikcgg:
### Summary
Add a new use case to the recording_service_recording_partitions_in_groups example.

### Details and comments
This demonstrates the use of <use_original_partitions> in RTI Replay Service, which controls whether or not the Replay tool should use the DDS Partition QoS information from the recorded discovery information data as it was originally received.


iblancasa:
@guyikcgg, is this something that can be merged in newer ConnextDDS releases?

@fgaranda, do we need to "port" these changes to `develop`?

guyikcgg:
@iblancasa, I haven't tested this example with newer versions of Recording Service. My understanding is that substantial changes were made to this service, so I think the example needs to be ported.

fgaranda:
Yes, this example should only target the 5.3.0 branch.

Issue #205, opened 2019-05-27
Add cmake script to build examples in all the forms.


lulivi:
Add cmake script to build examples in all `Release`/`Debug`-`Static`/`Dynamic` configurations.

lulivi:
Merged.

PR #208, opened 2019-05-29
[WIP] Add static analysis check with travis-ci

lulivi:
### Summary

Integrate clang-tidy static analysis with examples compilation using [`cmake` integration](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_CLANG_TIDY.html).


lulivi:
I'll add an argument in #205 script to allow passing the `CLANG-TIDY` variable value to `examples/connext_dds/CMakeLists.txt`.

resources/travis/linux_install.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class ZipFileWithPermissions(ZipFile):
+    """Custom ZipFile class handling file permissions.
+
+    Code from https://stackoverflow.com/a/54748564
+    """
+
+    def _extract_member(self, member, targetpath, pwd):
+
```
fgaranda: Can you double check the format of this file?

resources/travis/linux_script.py
```diff
@@ -0,0 +1,94 @@
...
+
+        print("Analyzing the build...", flush=True)
+        time_analysis_start = time.perf_counter()
+
+        result = sultan.analyze__build(
+            "--verbose",
+            "--status-bugs",
+            "--exclude",
+            Path.home().joinpath("rti_connext_dds-6.0.0/include"),
+            "--exclude",
```
fgaranda: Is there a way to avoid hardcoding the version here? Maybe we can ask for the path to rti_connext_dds-x.y.z, which would solve the problem of having to pass a specific DDS version here.

resources/travis/linux_install.py
lulivi: I'm not sure what you mean...

PR #215, opened 2019-06-14
[WIP] Add logging to find_package

iblancasa:
### Summary
CMake 3.15 added a cool new feature ([release notes](https://cmake.org/cmake/help/v3.15/release/3.15.html)):
> The message() command learned new types: NOTICE, VERBOSE, DEBUG and TRACE.

I am using this new feature in the FindPackage script for ConnextDDS.

### Details and comments
I implemented a function in order to wrapper the functionality. Why? This is something new in CMake 3.15 but I can make it work on previous releases (or something similar) using the macro and some logic.

The idea is to make easier to debug issues when we are looking for our libraries.

### DIFF
--- a/resources/cmake/FindRTIConnextDDS.cmake
+++ b/resources/cmake/FindRTIConnextDDS.cmake
@@ -275,6 +275,20 @@ function(connextdds_verbose message)
     message(VERBOSE ${message})
 endfunction()
 
+
+function(connextdds_log_xml XML_NAME XML)
+    string(REPLACE "\n"
+           ";" lines
+           ${XML})
+
+    message(VERBOSE "~~~~~~~START ${XML_NAME}~~~~~~~")
+    foreach(line ${lines})
+        message(VERBOSE "\t${line}")
+    endforeach()
+    message(VERBOSE "~~~~~~~FINISH ${XML_NAME}~~~~~~~")
+
+endfunction()
+
 #####################################################################
 # Preconditions Check                                               #
 #####################################################################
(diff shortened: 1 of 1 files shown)


iblancasa:
@jpovedano, @fgaranda, @lulivi: this feature is not finished but... feedback? Something interesting we can do with this?

resources/cmake/FindRTIConnextDDS.cmake
```diff
@@ -264,6 +264,31 @@
 # - All the Darwin platforms (OS X 10.11-10.13)
 #
 
+#####################################################################
+# Logging Functions                                                  #
+#####################################################################
+function(connextdds_debug message)
+    message(DEBUG ${message})
```
jpovedano: According to the documentation the DEBUG and VERBOSE levels were introduced in 3.15. Have we tried the behavior on previous versions? Were just undocumented levels?

jpovedano:
I like it. Some ideas we may want to explore:

- Make the logging functions a macro, so we can make use of CMAKE_CURRENT_LIST_LINE for better debuggability if desired
- Add our own prefix to DEBUG/VERBOSE so messages are easier to fllter

resources/cmake/FindRTIConnextDDS.cmake
```diff
@@ -264,6 +264,31 @@
...
+endfunction()
+
+function(connextdds_verbose message)
+    message(VERBOSE ${message})
```
fgaranda: You could rename these functions to:
connextdds_log_verbose
connextdds_log_debug

or

connextdds_message_verbose,
connextdds_message_debug

to give a context about what they do.

fgaranda:
Looks good! I have added a suggestion regarding the name, but other than that is pretty good.

resources/cmake/FindRTIConnextDDS.cmake
lulivi: I would go with `connextdds_log_xxxx` to match the function below (`connextdds_log_xml`)

PR #264, opened 2019-11-16
Improve flat_data_api example to show optionality

alexcamposruiz:
### Summary
Small changes to the publisher and subscriber to show that all members in an optional flatdata type are optional

### Details and comments
The publisher now omits the resolution field in some samples to demonstrate that all fields are optional. The subscriber now prints the resolution field if it exits. 

In the C++98 example, the subscriber now checks if the source member is present. Before, if it wasn't present the example may have crashed. In the C++11 example that's not necessary; if the source field doesn't exist, the example fails with an exception.


lulivi:
Hey @alexcamposruiz would you like to change the formatting issues found by clang-format? Checkout the [Travis log](https://travis-ci.org/rticommunity/rticonnextdds-examples/builds/612646950#L3226). You can find the changes requested in a diff-style format.

PR #271, opened 2019-11-20
Feature/record 1119

juanlu-rti:
### Summary

See [RECORD-1119](http://jira:8085/browse/RECORD-1119).

### Details and comments

The idea is to make sure we provide a note for users to use the `--config` switch with `cmake` when they're using multi-configuration generators. Also making sure we document the need to set the `LD_LIBRARY_PATH` variable when running the `service_as_lib` example.


iblancasa:
It looks good to me. Please, could you fix the warnings reported by Travis?

PR #293, opened 2020-04-15
Recording Service LevelDB integration example

juanlu-rti:

### Summary

This PR merges my work for the example showing integration with a NoSQL technology in Recording Service. See `README.md` file for more details.


examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,180 @@
...
+if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+    string(REGEX REPLACE "/GR" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /GR-")
+else()
+    string(REGEX REPLACE "-frtti" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti")
+endif()
+
+# Enable sanitizers
+if(ENABLE_SANITIZERS)
```
titosanxez, line 46: @iblancasa what do you think about this? is that something we should embed somewhere else? if not, is this the right way to do it?

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+This example is an implementation of the pluggable *RTI Recording Service* 
+storage, using the no-SQL, persistent key-value store, LevelDB. This engine was
+developed by Google as a fast key-value storage library that maps from key 
+strings to string values. The LevelDB code can be found on 
+[GitHub](https://github.com/google/leveldb).
+
+This example is written in modern C++ (C++11) and shows how integration with a
+no-SQL storage technology can be done in *RTI Recording Service*. The example
+implements both the Storage Writer and Storage Reader plug-ins, so it can be
+used to record, replay and/or convert the data coming from the Connext DDS data
```
titosanxez, line 14: add comma after replay

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+- `Tests.cxx` implements a test application that can be built and run on your
+  system to verify the correct working of the provided Storage Writer and 
+  Storage Reader classes.
+- `LevelDb_RecorderTypes.idl` provides the definitions of the types that will be
+  stored into the database keys and values. The code related to these types will
+  be automatically generated by the build system using *RTI Code Generator*.
+- `Utils.hpp` and `Utils.cxx` implement some helper functions and, more
+  importantly, the `UserKeyComparator` class that is used to customise the
+  sample ordering in LevelDB. Because *Recorder* and *Replay* work based on time
+  (they can be though about as time series storage) we need to make sure that
```
titosanxez, line 45: can be thought of

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+  samples are ordered in reception timestamp order, ascending and monotonic.
```
titosanxez, line 46: add comman after 'ascending'

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+converting) concurrently is _not supported_. In general, concurrently accessing
+the LevelDB database being recorded, replayed or converted is not supported in 
+this example. Note that this is not the case with our built-in SQLite 
+implementation, although concurrent access in that case is also discouraged.
+
+This example will store any DDS Topic discovered by *RTI Recording Service* that
+has been marked as to be recorded (by defining the Topics and Topic Groups in 
+the XML configuration).
+
+The example provides the following files:
```
titosanxez, line 26: All the content up to this point should be in a README.rmd that goes to the top-level directory (leveldb-integration). Then under cpp, we just leave the readme with specific content for this language, like the files description and build instructions

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+
+First of all, you need to make sure that the `LevelDbWriter` shared library
+is in the system's library path.  On Mac OS X, you must use the variable
+`RTI_LD_LIBRARY_PATH` instead of the common `DYLD_LIBRARY_PATH`.
+
+This example will work with any DDS Topic discovered on the DDS Domain ID of 
+your choice. Run any application publishing data into that domain, for example
+*RTI Shapes Demo* or *RTI DDS Ping*.
+
+The supplied XML configuration for *Recorder* expects you to set an environment
```
titosanxez, line 172: I would not use the enviornment word here, since the variable can be set with multiple ways. I recommend setting using it the -D command-line option

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+```bash
+export LEVELDB_WORKING_DIR=<storage location>
+cd build
+<connext dir>/bin/rtirecordingservice 
+        -cfgFile ../leveldb_recorder.xml
+        -cfgName LevelDbIntegration 
+        -domainIdBase <your chosen domain ID>
+```
+
+Note that if you don't want to export the necessary environment variables
```
titosanxez, line 204: I would switch this. Promote the -D as the way to do it, and do a note indicating that the enviornment can be used if preferred.
I find setting environment rather annoying, so I'd like to promote more an easier way to use variables

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+
+```bash
+RTI Recording Service (Recorder) 6.0.1 starting...
+RTI Recording Service started
+```
+
+The following directories will be created by *Recorder* after creating the
+Storage Writer instance:
+- `metadata.dat` database directory. This database will contain two key-value 
+  entries. They represent the nanoseconds timestamps where the LevelDB Storage
```
titosanxez, line 228: timestamps in nanoseconds

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+  discovered.
+- For every discovered and matched (allowed by the XML configuration) DDS Topic,
+  *Recorder* will create a database whose name will be defined as 
+  `<topic-name>@<domain-id>`. For example, if topic `Square` is discovered
+  on domain 54 (this is the Domain ID in the XML configuration, not the one
+  achieved by offsetting with the `-domainIdBase` command-line parameter), 
+  then *Recorder* will create database (directory) `Square@54` and will store
+  all Square samples received in that domain in that database.
+
+## Format of the stored data
```
titosanxez, line 245: this seems to be a section for the geernal overview readme

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+```
+
+The `UserDataKey` type is used to store keys for both publication data and user
+data samples.
+
+The LevelDB library allows for the customization of the key ordering mechanism.
+This example implements a key comparator class that extends LevelDB's 
+`Comparator` interface. This implementation achieves ordering of samples in a
+strictly time-based manner. Resulting databases order their samples in an
+ascending, monotinic order based on the reception timestamp. See class
```
titosanxez, line 275: monotonic

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+([Format of the stored data](#format-of-the-stored-data)). The code
+for the types defined in file `LevelDb_RecorderTypes.idl` will be
+generated automatically. The generated code includes methods to
+serialize and deserialize samples to and from CDR format.
+
+The main members of the `LevelDbStreamWriter` class are:
+
+- `data_db_`: this is the LevelDB database object (leveldb::DB) that 
+  will be used to store the data.
+- `key_comparator`: the custom key comparator will be passed when
```
titosanxez, line 461: key_comparator_ ?

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+```
+
+In the case of the value buffer, the serialized sample's max size
+not only depends on the size of type `UserDataValue`, but mostly
+on the size of the topic's type's sample max serialized size. Hence,
+we first get the type from the `StreamInfo` object, which is
+represented as a native `DDS_TypeCode` pointer object. With it, we
+do the following:
+
+- If the type is unbounded, we initialize the buffer to 8 KB.
```
titosanxez, line 499: why not doing always this? potentially we could have a default of, perhaps, 64KB and resize as needed. Even we could mention that this minimum configuraiton could be configurable if needed.

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+      fields are available in the `SampleInfo` object.
+      
+    - For every sample, prepare a **value** entry: get the valid data
+      flag and store it in the value object. If the sample is valid,
+      serialize the sample into CDR. Here is the code snippet:
+        ```cpp
+        UserDataValue value;
+        value.valid_data(sample_info.valid());
+        if (sample_info.valid()) {
+            const dds::core::xtypes::DynamicData& sample = *(sample_seq[i]);
```
titosanxez, line 548: why this line? why not passing *sample_seq[i] directly to_cdr_buffer()?

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+`ReducedDCPSPublication` is used as the value data. The code
+for the types defined in file `LevelDb_RecorderTypes.idl` will be
+generated automatically. The generated code includes methods to
+serialize and deserialize samples to and from CDR format.
+
+The main members of the `PubDiscoveryLevelDbWriter` class are:
+
+- `discovery_db_`: this is the LevelDB database object (leveldb::DB)
+  that will be used to store the `DCPSPublication`.
+- `key_comparator`: the custom key comparator will be passed when
```
titosanxez, line 578: key_comparator_?

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+
+The main members of the `LevelDbStreamReader` class are:
+
+- `data_db_`: this is the LevelDB database object (`leveldb::DB`) that 
+  will be used to read the data and that was recorded by *Recorder* using
+  the Storage Writer implementation in this example.
+- `db_iterator_`: the class keeps a database iterator (`leveldb::Iterator`)
+  that points to the next sample to be read.
+- `current_timestamp_`: the current sample's reception timestamp.
+- `key_comparator`: the custom key comparator will be passed when
```
titosanxez, line 759: this seems to be a copy-carry error (assuming it is)

examples/recording_service/leveldb-integration/cpp/README.md
```diff
@@ -0,0 +1,1077 @@
...
+    } else {
+        next_sample_valid = true;
+    }
+} else {
+    next_sample_valid = false;
+}
+```
+
+Notice that we need to check if the next sample, if any,
+is within the stream rea'ders time range (`current_timestamp`
```
titosanxez, line 914: reader's

examples/recording_service/leveldb-integration/cpp/Utils.cxx
```diff
@@ -0,0 +1,48 @@
...
+UserDataKeyComparator::UserDataKeyComparator() = default;
+
+UserDataKeyComparator::~UserDataKeyComparator() = default;
+
+
+int UserDataKeyComparator::Compare(
+        const leveldb::Slice& a,
+        const leveldb::Slice& b) const
+{
+    return memcmp(a.data(), b.data(), a.size());
```
titosanxez, line 31: this implementation does not provide time increasing monotonic accross all samples, but just within the same DW correct? different samples for different DW can be out of time order

examples/recording_service/leveldb-integration/cpp/leveldb_converter.xml
```diff
@@ -0,0 +1,82 @@
...
+        </domain_participant>
+        <session name="Session">
+            <topic_group name="AllExceptRti" participant_ref="Domain0">
+                <allow_topic_name_filter>*</allow_topic_name_filter>
+                <deny_topic_name_filter>rti/*</deny_topic_name_filter>
+            </topic_group>
+        </session>
+    </converter>
+    
+    <converter name="LevelDb_To_SQLiteJSON">
```
titosanxez, line 49: this is pretty cool

examples/recording_service/leveldb-integration/cpp/LevelDbWriter.cxx
```diff
@@ -0,0 +1,512 @@
...
+            leveldb::WriteOptions(),
+            end_time_key_name(),
+            current_time_stream.str());
+    if (!db_status.ok()) {
+        std::stringstream log_stream;
+        log_stream << "Failed to write end time to metadata DB" << std::endl;
+        log_stream << "    LevelDB error = " << db_status.ToString()
+                << std::endl;
+        /* Do not throw in a C++ destructor */
+        std::cerr << log_stream.str();
```
titosanxez, line 148: just as a note: starting next version we'll be able to leverage ROUTING-612.

examples/recording_service/leveldb-integration/cpp/LevelDbWriter.cxx
```diff
@@ -0,0 +1,512 @@
...
+        /*
+         * The user-data sample is stored along with valid data flag. In fact,
+         * if the sample is not valid we do not serialize the sample to a CDR
+         * buffer at all
+         */
+        UserDataValue value;
+        value.valid_data(sample_info.valid());
+        if (sample_info.valid()) {
+            const dds::core::xtypes::DynamicData& sample = *(sample_seq[i]);
+            rti::core::xtypes::to_cdr_buffer(value.data_blob(), sample);
```
titosanxez, line 309: this operation will resize data_blob() as needed, which is fine for the example. Then, why bother with the unbounded situation for value_buffer_?

examples/recording_service/leveldb-integration/cpp/LevelDbReader.cxx
```diff
@@ -0,0 +1,637 @@
...
+                &ex);
+        if (ex != DDS_NO_EXCEPTION_CODE) {
+            std::stringstream log_stream;
+            log_stream << "Could not get serialized sample max size for stream"
+                    << std::endl;
+            log_stream << "    Stream name = " << stream_info.stream_name()
+                    << std::endl;
+            throw std::runtime_error(log_stream.str());
+        }
+        value_buffer_.resize((std::vector<char>::size_type) size);
```
titosanxez, line 225: same comment than for the writer part. Perhaps is not worth doing any of this, in any case we're allocating for value.blob()

examples/recording_service/leveldb-integration/cpp/LevelDbReader.cxx
```diff
@@ -0,0 +1,637 @@
...
+    }
+    /*
+     * Fill out the collections passed as parameters by the service. We will use
+     * the internal vectors of samples and sample infos to get pointers
+     * to be stored in these collections
+     */
+    sample_seq.resize(num_samples_read);
+    info_seq.resize(num_samples_read);
+    for (int32_t i = 0; i < num_samples_read; i++) {
+        sample_seq[i] = loaned_samples_[i].get();
```
titosanxez, line 316: I would indicate this works because Replay guarantees that for each read() and return_loan() follows, and hence we can use a single cached pool of samples

examples/recording_service/leveldb-integration/cpp/LevelDbReader.cxx
```diff
@@ -0,0 +1,637 @@
...
+
+LevelDbStreamInfoReader::~LevelDbStreamInfoReader()
+{
+}
+
+void LevelDbStreamInfoReader::read(
+        std::vector<rti::routing::StreamInfo *>& sample_seq,
+        const rti::recording::storage::SelectorState& selector)
+{
+    if (finished_) {
```
titosanxez, line 481: why is not Replay doing this independently of the plugin?

examples/recording_service/leveldb-integration/cpp/LevelDbReader.cxx
```diff
@@ -0,0 +1,637 @@
...
+                        "(will be skipped)" << std::endl;
+                log_stream << "    Stream name = "
+                        << reduced_sample.topic_name() << std::endl;
+                std::cerr << log_stream.str();
+                continue;
+            }
+            stream_info->type_info().type_representation(type_code);
+        }
+        loaned_stream_infos_.resize(num_samples_read + 1);
+        loaned_stream_infos_[num_samples_read] = stream_info;
```
titosanxez, line 544: this making an assignment of the stream_info shared ptr correcT?

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,180 @@
...
+    RTIConnextDDS "6.0.1"
+    REQUIRED
+    COMPONENTS
+        core
+        routing_service
+)
+
+set(LEVELDB_LIB_NAME "leveldb")
+
+find_library(
```
iblancasa, line 30: I downloaded and installed `LevelDB` and it has integration with CMake.
The way to activate it is adding the folder `<where LevelDB is installed>/lib64/cmake/leveldb/` to `CMAKE_MODULE_PATH` and you will be able to find it using `find_package`.


So the problem now: how can we find the folder where LevelDB is installed? Well, we can ask the user to provide it using the `LEVELDB_DIR` CMake variable but we can try to add some logic to try to find it... Like the default path where it is installed in Windows, Linux and Darwin.
Also, since we know the library will be installed under `lib64`, we can run `find_library` to find the library (without specifying any path to ask to CMake to search on its default paths) and them add `<where the library is placed>/cmake/leveldb/` to `CMAKE_MODULE_PATH` to be able to use `find_package`. Why can this be useful? If, in the future, they decide to add compiler flags, compiler definitions, dependencies with other libraries or other things, that information will be included in their CMake scripts and our `CMakeLists.txt` file will still be valid.

examples/recording_service/leveldb-integration/cpp/LevelDbReader.hpp
```diff
@@ -0,0 +1,262 @@
...
+
+    virtual bool finished();
+
+    virtual void reset();
+
+private:
+
+    std::unique_ptr<leveldb::DB> metadata_db_;
+
+    int64_t service_start_time_;
```
titosanxez, line 232: at some point we need to create this as a type we ship (along wiht other things like SN, etc). Is this something part of the reusability ideas you have?

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
titosanxez: Thank you Israel for the thorough review of the build file.

In a separate note, I can anticipate people will do the same things differently not followiing our methods. We should think about a creating some sort of template (e..g, a velocity one to be used with rtiddsgen) so that all of these things can be reused. We could also have a simple tool added at the top-level of the repository to generate the CMake.

Just leaving the thought here. We can discuss further how to do it if you agree

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
juanlu-rti: It is

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: Done

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: Done.

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: Done

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: Will do last thing.

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: Yes, that's true. I do prefer the -D command-line option too.

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: I've switched the order and the text so that `-D` is promoted

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: Done

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: Done

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: Yes

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: It's an option. But if we can reserve memory upfront during startup, why doing it afterwards? That was my thinking.

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: For clarity, for the users reading the code and the doc.

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: Yes he he

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: I just corrected all occurrences.

examples/recording_service/leveldb-integration/cpp/README.md
juanlu-rti: Done

examples/recording_service/leveldb-integration/cpp/Utils.cxx
juanlu-rti: Yes, that's right

examples/recording_service/leveldb-integration/cpp/LevelDbWriter.cxx
juanlu-rti: I have just created https://jira.rti.com/browse/RECORD-1163 to track this.

examples/recording_service/leveldb-integration/cpp/leveldb_converter.xml
juanlu-rti: ^^

examples/recording_service/leveldb-integration/cpp/LevelDbWriter.cxx
juanlu-rti: In the bounded case, we will avoid memory allocations in the critical path, by setting the buffer's capacity to the max serialized sample's size.

examples/recording_service/leveldb-integration/cpp/LevelDbReader.cxx
juanlu-rti: We could go for the simplest case, although I prefer to avoid the allocations in the critical path for the bounded case.
But if you prefer it, I can change it, I'm not too attached to the solution.

examples/recording_service/leveldb-integration/cpp/LevelDbReader.cxx
juanlu-rti: Done

examples/recording_service/leveldb-integration/cpp/LevelDbReader.cxx
juanlu-rti: I think I documented this in the issue that I created with the common behavioural patterns to implement as helpers.

examples/recording_service/leveldb-integration/cpp/LevelDbReader.cxx
juanlu-rti: Yes, that is the case.

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,180 @@
...
+    leveldb_lib
+    NAMES ${LEVELDB_LIB_NAME}
+    PATHS "${LEVELDB_DIR}"
+)
+
+# Disable RTTI.
```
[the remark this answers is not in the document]
juanlu-rti: Nope. This is actually RTTI, Real-Time Type Information system in C++.

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,180 @@
...
+if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
```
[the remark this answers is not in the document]
juanlu-rti: Well, if I recall correctly: RTTI is enabled by default. But LevelDB needs to disable it, and we have to disable it too to be able to load the LevelDB libraries within our plugin. Hence, if we detect the `-frtti` flag we remove it. And after that, we add the flag to actually disable it (`-fno-rtti`)

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,180 @@
...
+    string(REGEX REPLACE "/GR" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /GR-")
+else()
+    string(REGEX REPLACE "-frtti" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti")
+endif()
+
+# Enable sanitizers
+if(ENABLE_SANITIZERS)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address  -fsanitize=leak -g")
```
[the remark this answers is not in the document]
juanlu-rti: Shouldn't we use `add_compile_options` instead?

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,180 @@
...
+add_custom_target(
+    generated_code 
+    DEPENDS 
+        ${REDUCED_DCPSPUB_CPP_SOURCES}
+        generated_code_dir
+)
+
+# Define the library that will provide the storage writer plugin
+add_library(
+    LevelDbWriter MODULE
```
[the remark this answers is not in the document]
juanlu-rti: So, I should use SHARED here, right?

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,180 @@
...
+    "${CMAKE_CURRENT_BINARY_DIR}/generated/LevelDb_RecorderTypes.cxx"
```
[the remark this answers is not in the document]
juanlu-rti: Absolutely. I've also renamed the variable to `RECORDER_TYPES_CPP_SOURCES`.

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,180 @@
...
+    PROPERTIES CXX_STANDARD 11
+)
+
+
+target_link_libraries(
+    LevelDbWriter
+    RTIConnextDDS::routing_service_infrastructure
+    RTIConnextDDS::cpp2_api
+    "${leveldb_lib}"
+    ${CONNEXTDDS_EXTERNAL_LIBS}
```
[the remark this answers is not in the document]
juanlu-rti: Ah, you may be right. I've added it by imitation to what we do in other examples. I'll tell you once I test this.

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
juanlu-rti: One note about LevelDB: while its CMakeLists.txt is ready to build as shared library (we need that), the bundle doesn't provide that by default, if I recall correctly.
@iblancasa , when you said you installed levelDB, you mean you used an installation bundle or just unzipped something from github?

examples/recording_service/leveldb-integration/cpp/LevelDbReader.hpp
juanlu-rti: What type are your referring to?

examples/recording_service/leveldb-integration/cpp/README.md
titosanxez: I agree, but I was considering that we could always follow that approach for simplicity in the code. It's not a big deal.

examples/recording_service/leveldb-integration/cpp/LevelDbReader.hpp
titosanxez: I was thinking to potentially create a struct that contains all these members, if these are going to be appearing in the implementation anyway. Something like ReplayTimeProperty or something along those lines

examples/recording_service/leveldb-integration/cpp/LevelDbReader.cxx
titosanxez: I'm fine with it.

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
iblancasa: @juanlu-rti I downloaded the source code, built it and installed the result. Also, I tried with the package from the Fedora repositories.

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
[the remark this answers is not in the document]
iblancasa: Oh. I see. Thanks!

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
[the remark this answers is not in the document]
iblancasa: Oh, I see. CMake is adding always `/GR`... Ok, we can look for a better solution in the meantime but it looks good for me. Thanks!

examples/recording_service/leveldb-integration/cpp/CMakeLists.txt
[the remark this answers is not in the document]
iblancasa: `add_compile_options` will apply globally to all the project. If we include this folder as part of another, all the targets will be affected by the options added in `add_compile_options`.

Using `target_compile_definitions`, you ensure your changes just applies for the desired target :)

Antoniolm (approved):

alexcamposruiz:
This PR hasn't been updated in a long time.

PR #294, opened 2020-04-21
Support multiconfiguration projects and other fixes

iblancasa:
### Summary
What is included in this PR?
* Support for multi-configuration projects
* Read the version from the header files. No more dependency on the `rti_versions.xml` file. The check of the `rti_versions.xml` can be enabled with a variable
* Fixes in the imported targets for libraries like `metp`, `Routing Service` or the low bandwidth plugins
* Improved the way we try to guess the `CONNEXTDDS_DIR` folder when is not provided
* Fail if a version older than CMake 3.11 is used
* Use the imported targets to links against OpenSSL

### DIFF
--- a/examples/connext_dds/CMakeLists.txt
+++ b/examples/connext_dds/CMakeLists.txt
@@ -3,7 +3,7 @@
 # without express written permission.  Any such copies, or revisions thereof,
 # must display this notice unaltered.
 # This code contains trade secrets of Real-Time Innovations, Inc.
-cmake_minimum_required(VERSION 3.7)
+cmake_minimum_required(VERSION 3.11)
 project(rtiexamples-connext-dds)
 set(CMAKE_MODULE_PATH
     ${CMAKE_MODULE_PATH}
@@ -11,7 +11,8 @@ set(CMAKE_MODULE_PATH
 )
 
 find_package(RTIConnextDDS
-    "6.0.0"
+    "6.0.1"
+    EXACT
     REQUIRED
     COMPONENTS
         core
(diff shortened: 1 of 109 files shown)


examples/connext_dds/CMakeLists.txt
```diff
@@ -3,15 +3,16 @@
...
+cmake_minimum_required(VERSION 3.11)
 project(rtiexamples-connext-dds)
 set(CMAKE_MODULE_PATH
     ${CMAKE_MODULE_PATH}
     "${CMAKE_CURRENT_SOURCE_DIR}/../../resources/cmake"
 )
 
 find_package(RTIConnextDDS
-    "6.0.0"
+    "6.0.1"
```
iblancasa, line 14: Revert this

resources/cmake/FindRTIConnextDDS.cmake
```diff
@@ -27,10 +27,10 @@
...
 # libraries associated with them.
 #
-# Also, the version consistency across all the requested components will be
+# Also, the version consistency across all the requested components can be
 # checked. If the version between the components missmatch,
-# `RTIConnextDDS_FOUND` will be set to `FALSE`. This feature can be disabled
-# setting `CONNEXTDDS_DISABLE_VERSION_CHECK` to `TRUE`.
+# `RTIConnextDDS_FOUND` will be set to `FALSE`. This feature can be enabled
+# setting `CONNEXTDDS_ENABLE_VERSION_CHECK` to `TRUE`.
 #
```
fgaranda, line 34: Maybe we want to call this ENABLE_VERSION_CONSISTENCY_CHECK. If I understand this correctly, we always check we're using the right version using a header file, but this check is actually going a bit further and parsing rti_versions.xml, right? Maybe we should be a bit more explict about it.

fgaranda (approved):

PR #300, opened 2020-06-02
Move from markdownlint-js to mdl ruby.

### LINKED ISSUE #299: Change markdown linter
### Information

It seems that the markdown JS linter is failing since some days. We are going to change it to mdl Ruby package, wich seems more stable.

lulivi:
Closes #299


.travis.yml
```diff
@@ -11,7 +11,7 @@ addons:
     update: true
 before_install:
   - git config --global clangFormat.binary clang-format-9
-  - npm install -g markdownlint-cli
+  - gem install mdl
```
iblancasa, line 14: Could we force an specific version for `mld`?

lulivi: For sure!

PR #305, opened 2020-06-05
Move md and py lint to GitHub Actions

lulivi:
### Summary

Move Markdown and Python lint to GitHub Actions.

### Details and comments

- Moved Markdown and Python formatting checks to GitHub Actions for its greater configuration.
- Reordered Travis steps to check C/C++ files format before the static analysis (to avoid getting a format error after the long static analysis wait).
- Moved back to `markdownlint-cli` as the Markdown linter due to error hints.

PR #306, opened 2020-06-11
Move md and py lint to GitHub Actions (#305)

lulivi:
Move markdown and python linting to github actions


lulivi:
First we will merge master in develop

PR #313, opened 2020-10-26
MongoDB Adapter Example

titosanxez:


examples/routing_service/routing_service_mongo_db/cpp/RsMongoGateway.xml
```diff
@@ -0,0 +1,115 @@
+<?xml version="1.0"?>
```
fgaranda, line 1: Can you add a copyright notice to this file?

fgaranda (changes requested):

PR #314, opened 2020-10-28
Update waitsets_subscriber.cs

schlesg:
The wrong index was used to print the sample.


fgaranda:
Looks good Gal! Could you sign the CLA https://community.rti.com/cla before we merge the PR?

PR #339, opened 2021-02-26
Add C# instance_statistics example

alexcamposruiz:
### Summary
Add C# directory for the instance_statistics example

### Details and comments
* This example uses the new C# API.
* I used better names for the IDL file, namespace, and type. I think the other examples should be updated.


samuelraeburn (approved):

softwarejimenez (approved):

examples/connext_dds/instance_statistics/cs/InstanceStatisticsExampleProgram.cs
```diff
@@ -0,0 +1,161 @@
+/*
+* (c) Copyright, Real-Time Innovations, 2021.  All rights reserved.
```
fgaranda, line 2: These * are off by one space I think, same happens with other files.

fgaranda (changes requested):

fgaranda (approved):

PR #340, opened 2021-03-01
Real-Time WAN Transport example

fercsrti77:
### Summary

New Real-Time WAN Transport Example

### Details and comments

A few things to notice:
* The example can be run with and without security
* The example can be run with and without Cloud Discovery Service

The integration with security required certificates, permissions, and governance files. I have followed the same model we use for the shipped examples.

There is a new directory called **dds_security** under **examples** containing the permissions and certificates. These files are supposed to be shared by all the applications testing security. At this point it is only the real_time_wan_transport.

I have updated **resources/cmake/ConnextDdsAddExample.cmake** to allow the deployment of the security-related files in the binary directories for the applications. If an application uses security you must called connextdds_add_example with the option REQUIRE_SECURITY.

@fgaranda I have made my changes in develop. I could not find in the documentation the starting branch. If it has to be different let me know.


examples/connext_dds/real_time_wan_transport/c++/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+# Include ConnextDdsAddExample.cmake from resources/cmake
+include(ConnextDdsAddExample)
+
+connextdds_add_example(
+    IDL "real_time_wan_transport"
+    LANG "C++"
+    REQUIRE_SECURITY TRUE
+)
+
+configure_file(
```
fercsrti77, line 22: Not sure what is the best way to deploy the **CLOUD_DISCOVERY_SERVICE.xml** configuration. I took this from the way we deploy Routing Service files. However, another option is to assume that some of the examples may ship with some IS configuration and in such case we could integrate the copy inside connextdds_add_example

examples/connext_dds/real_time_wan_transport/c++/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+    "${CMAKE_CURRENT_SOURCE_DIR}/../../../../resources/cmake"
+)
+
+# Include ConnextDdsAddExample.cmake from resources/cmake
+include(ConnextDdsAddExample)
+
+connextdds_add_example(
+    IDL "real_time_wan_transport"
+    LANG "C++"
+    REQUIRE_SECURITY TRUE
```
fercsrti77, line 19: When this option is present the dds_security directory containing certificates, permissions, and governance will be copied under the application binary directory.

examples/connext_dds/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ set(CMAKE_MODULE_PATH
 )
 
 find_package(RTIConnextDDS
-    "6.0.0"
+    "6.1.0"
```
fercsrti77, line 14: This example only works with 6.1.0. This is why I changed the requirements for RTIConnextDDS

fgaranda (approved):

PR #352, opened 2021-03-08
C#: new examples and warning about deprecated API

alexcamposruiz:
### Summary
Update for the new C# API. Port the examples of multichannel, partitions. Add warning about deprecated API

### Details and comments
The multichannel and partitions examples have been ported to the new C# API. The old code has been deleted.

The rest of C# examples that use the old API now contain a warning in their README that points to the GSG.

Also: updated incorrect log message in all the partitions examples. The match expression is "A, D", not "AD".


examples/connext_dds/multichannel/cs/MarketDataSubscriber.cs
```diff
@@ -0,0 +1,112 @@
+/*
+* (c) Copyright, Real-Time Innovations, 2012.  All rights reserved.
```
iblancasa, line 2: I think you need to fix here the copyright year

examples/connext_dds/partitions/cs/PartitionsExampleSubscriber.cs
```diff
@@ -0,0 +1,104 @@
+/*
+* (c) Copyright, Real-Time Innovations, 2012.  All rights reserved.
```
iblancasa, line 2: I think you need to fix here the copyright year

iblancasa (changes requested):

iblancasa (approved):

examples/connext_dds/asynchronous_publication/cs/README.md
```diff
@@ -1,5 +1,9 @@
 # Example Code: Asynchronous Publication
 
+:warning: **Note**: This example uses the **Legacy .NET API**. Until it is
+updated, you can learn more about the **new Connext DDS C# API** in the
+[Getting Started Guide](https://community.rti.com/static/documentation/connext-dds/current/doc/manuals/connext_dds/getting_started/index.html).
```
fgaranda, line 5: You could use https://community.rti.com/static/documentation/connext-dds/6.1.0/doc/manuals/connext_dds_professional/getting_started_guide/csharp/before_csharp.html instead.

We can always do a bulk change when we bump the version, but the link to current is going to break.

fgaranda (changes requested):

samuelraeburn (approved):
Just one comment on possibly missing file, if it isn't missing all looks fine.

examples/connext_dds/multichannel/cs/MarketDataSubscriber.cs
```diff
@@ -0,0 +1,112 @@
+/*
```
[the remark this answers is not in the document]
alexcamposruiz: MarketDataProgram.cs is generated by rtiddsgen and doesn't need to change. I've updated the README because the codegen instructions were wrong.

fgaranda (approved):

PR #375, opened 2021-03-23
C# DynamicData pub/sub example

alexcamposruiz:
### Summary
Port DynamicData pub/sub example to new c# API

### Details and comments
The example follows the general idea of the other implementations, but adds an option to create the DynamicType in different ways, including an extended ShapeType (see README)


luisrg-rti (approved):
Looks good and easy to follow

PR #396, opened 2021-03-29
Port builtin topic example to new C# API

alexcamposruiz:
### Summary
Port builtin topic example to new C# API

### Details and comments
* The example roughly follows the previous example
* But it allows setting the password and the topic name in the command line.
* I also added a disclaimer to the common README that this is not a secure authentication method.


examples/connext_dds/builtin_topics/cs/BuiltinTopicsExamplePublisher.cs
```diff
@@ -0,0 +1,173 @@
...
+            using LoanedSamples<ParticipantBuiltinTopicData> samples = participantReader
+                .Select()
+                .WithState(ViewState.New.And(InstanceState.Alive))
+                .Take();
+
+            foreach (var sample in samples)
+            {
+                if (!sample.Info.ValidData)
+                {
+                    continue;
```
luisrg-rti, line 122: Maybe mention why this is necessary across this example code instead of ValidData()

luisrg-rti (changes requested):

examples/connext_dds/builtin_topics/cs/BuiltinTopicsExamplePublisher.cs
alexcamposruiz: I can't find an explanation that is short and doesn't distract from the example. I'll leave it as is, since other examples do show the ValidData() enumerator.

PR #399, opened 2021-04-06
Update clang-format configuration to check java source files

### LINKED ISSUE #398: Apply format and code conventions to java source files
## Enhacement description

To ensure the best practices in `Java` code it would be convenient to use a tool that allows us to format the source code and check if a file (or files) is following the style/format rules. In addition, it would be necessary to create a new workflow to check the code in each push or pull request.

ManuelJNunez:

### Summary

I updated the clang-format configuration so that it checks java source files too.

### Details and comments

close #398

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -10,8 +10,11 @@ BasedOnStyle: WebKit
 # Formatting rules in our coding conventions
 ###############################################################
 
-# CFR1-01: 80-column line
+# Languages
 Language: Cpp
+Language: Java
+
+# CFR1-01: 80-column line
 ColumnLimit: 80
 
 # CFR1-02: Tab character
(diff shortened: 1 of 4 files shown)


lulivi:
I'm thinking that we may want to separate cxx and java files from the formatting steps, right? What do you think @ManuelJNunez and @angelrti

ManuelJNunez:
@lulivi maybe I can use the separators in the same file. Take a look [here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).

.clang-format
```diff
@@ -138,3 +138,113 @@ BreakBeforeBinaryOperators: NonAssignment
...
+# CFR3-05: Statements
+AllowShortBlocksOnASingleLine: false
+AllowShortFunctionsOnASingleLine: false
+
+# CFR3-06: Pointer declaration
+# Checkout CFR1-08
+
+BreakBeforeBraces: Custom
+BraceWrapping:
+    AfterClass: false             # CFR2-03: C classes
```
lulivi, line 210:
```suggestion
    AfterClass: false            # CFR2-03: C classes
```

lulivi:
I'm not sure if we really want some duplicated code in the .clang-format file, because they are the same, right?

lulivi (approved):

resources/ci_cd/linux_format.py
```diff
@@ -97,7 +97,7 @@ def get_argument_parser() -> ArgumentParser:
         help="Disable markdownlint linter.",
     )
     parser.add_argument(
-        "-cxx",
+        "-clang",
```
angelrti, line 100: we need to update the contributing.md accordingly

angelrti (changes requested):

resources/ci_cd/linux_format.py
ManuelJNunez: Done 👍🏻

angelrti (approved):

lulivi (approved):

PR #449, opened 2021-06-22
Update C++98 examples template

### LINKED ISSUE #438: [ENH] Move from c++ to c++98 language directory names to homogenize the naming
### Enhancement description

The language directories names after this will be:
- c
- c++98
- c++11
- cs
- java

ManuelJNunez:
### Summary

This Pull Request is for updating the codegen template of C++98 examples.

### Details and comments

close #438 

### Checks


ManuelJNunez:
This is a draft, so it is not finished yet. Once I have finished the update of all the examples, I will convert it to a Pull Request.

examples/connext_dds/asynchronous_publication/c++98/application.h
```diff
@@ -0,0 +1,132 @@
+/*
```
fgaranda, line 1: If we're changing to the new templates. I think we don't need to check in application.h, because it'll be generated by rtiddsgen, similarly to the rest of cxx and h files.

fgaranda (changes requested):

examples/connext_dds/asynchronous_publication/c++98/application.h
ManuelJNunez: Actually, it is necessary. The CMake build infrastructure does not generate the application.h files, and some application.h has been modified because there are some examples with different or extra arguments.

jpovedano (changes requested):
Since most of the examples are pretty similar there are some issues that we would need to address in all the examples:

- New line in application.h
- Some `take()` calls are not capturing the return code
- Some `process_data` functions are not returning a value or returning the wrong value
- Some DDS_ReturnCode_t are not initialized
- Use `instance_handle` instead of using `DDS_HANDLE_NIL` when writing data. Since there is code to be un/commented to register the instance, uncomenting this code becomes useles

examples/connext_dds/custom_content_filter/c++98/ccf_subscriber.cxx
```diff
@@ -252,74 +122,116 @@ extern "C" int subscriber_main(int domainId, int sample_count)
...
-                return -1;
+                return shutdown_participant(
+                        participant,
+                        "set_expression_parameters error",
+                        EXIT_FAILURE);
             }
-        } else if (count == 20) {
-            printf("changing filter parameters\n");
-            printf("Filter: 3 divides x\n");
+        } else if (samples_read == 10) {
```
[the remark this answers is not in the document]
ManuelJNunez: Because it worked with loop iterations and now with samples_read. This is necessary to make it shorter to test with Valgrind.

PR #469, opened 2021-11-09
Fixed the custom transport example. Doing other than a non-symmetric …

franporcel:
…IP address did not work because the example assumed the endianness. Using memcpy instead of shifting bytes fixes it

### Checks

### DIFF
--- a/examples/connext_dds/custom_transport/c/FileTransport.c
+++ b/examples/connext_dds/custom_transport/c/FileTransport.c
@@ -1781,10 +1781,8 @@ RTI_INT32 NDDS_Transport_FILE_get_receive_interfaces_cEA(
     /* Set the meaningful bytes of the address */
     {
         struct in_addr ipAddr;
-        unsigned int ipAddrNetworkOrder;
 
         inet_aton(me->_property.address, &ipAddr);
-        ipAddrNetworkOrder = ipAddr.s_addr;
         memcpy(
                 interface_array_inout[0].address.network_ordered_value + 12,
                 &ipAddr,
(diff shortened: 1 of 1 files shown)


fgaranda (approved):

examples/connext_dds/custom_transport/c/FileTransport.c
```diff
@@ -1781,18 +1781,12 @@ RTI_INT32 NDDS_Transport_FILE_get_receive_interfaces_cEA(
...
-        ipAddrNetworkOrder = ipAddr.s_addr;
-        interface_array_inout[0].address.network_ordered_value[15] =
-                (ipAddrNetworkOrder) % 256;
-        interface_array_inout[0].address.network_ordered_value[14] =
-                (ipAddrNetworkOrder >> 8) % 256;
-        interface_array_inout[0].address.network_ordered_value[13] =
-                (ipAddrNetworkOrder >> 16) % 256;
-        interface_array_inout[0].address.network_ordered_value[12] =
-                (ipAddrNetworkOrder >> 24) % 256;
+        memcpy(
```
juanlu-rti, line 1786: The C/C++ linter is failing for this line. I guess the first parameter goes in the same line.

juanlu-rti, line 1786: The fix looks good to me 👍

fgaranda: That doesn't seem right. This is the right indentation, we'll merge it like this and I'll investigate.

PR #470, opened 2021-11-09
Update CLOUD_DISCOVERY_SERVICE.xml to lower refresh_period

rgrti:
### Summary
The default refresh_period is 60 seconds which can cause two participants to take a significant amount of time before data samples are exchanged.

### Details and comments
This change lowers the refresh period from the default 60 seconds down to 5 seconds to speed up initial startup

### Checks


rgrti:
I agree, [PR468](https://github.com/rticommunity/rticonnextdds-examples/pull/468) includes these changes and is more comprehensive in what it fixes. This PR470 is not needed anymore

Issue #476, opened 2021-11-18
[BUG] Compile and warning issues on /examples/recording_service/pluggable_storage/ 


Antoniolm:
### System information

- **RTI Product**: RTI Connext
- **Version**: 6.1.0
- **Operating system**: Windows
- **Compiler**: VS2015
- **Compiler version**:
- **Additional information**:

### What is the current behavior?
a) Several warnings on the compilation of the pluggable storage C example.
b) Compile and warning issues on pluggable storage C++ example

### Steps to reproduce the issue
Just run it out of the box on Windows.

PR #477, opened 2021-11-18
#476 Fixed compilation and warning issue on Recording Service pluggable example

Antoniolm:
### Summary
Fixed some warning and compilation issues on Recording Service pluggable example on Windows. 

### Checks

### DIFF
--- a/examples/recording_service/pluggable_storage/c/CMakeLists.txt
+++ b/examples/recording_service/pluggable_storage/c/CMakeLists.txt
@@ -83,6 +83,7 @@ add_library(
     FileStorageWriter
         "${CMAKE_BINARY_DIR}/generated/HelloMsg.c"
         "${CMAKE_CURRENT_SOURCE_DIR}/FileStorageWriter.c"
+        "${CMAKE_CURRENT_SOURCE_DIR}/FileStorageUtils.c"
 )
 
 add_dependencies(FileStorageWriter generated_code)
@@ -107,6 +108,7 @@ add_library(
     FileStorageReader
         "${CMAKE_BINARY_DIR}/generated/HelloMsg.c"
         "${CMAKE_CURRENT_SOURCE_DIR}/FileStorageReader.c"
+        "${CMAKE_CURRENT_SOURCE_DIR}/FileStorageUtils.c"
 )
 
 add_dependencies(FileStorageReader generated_code)
--- a/examples/recording_service/pluggable_storage/c/CMakeLists.txt
+++ b/examples/recording_service/pluggable_storage/c/CMakeLists.txt
@@ -84,6 +84,7 @@ add_library(
         "${CMAKE_BINARY_DIR}/generated/HelloMsg.c"
         "${CMAKE_CURRENT_SOURCE_DIR}/FileStorageWriter.c"
         "${CMAKE_CURRENT_SOURCE_DIR}/FileStorageUtils.c"
+        "${CMAKE_CURRENT_SOURCE_DIR}/FileStorageUtils.h"
 )
 
 add_dependencies(FileStorageWriter generated_code)
@@ -100,7 +101,8 @@ target_include_directories(
     PUBLIC
         "${CMAKE_CURRENT_BINARY_DIR}"
         "${CMAKE_BINARY_DIR}/generated"
-        "${CONNEXTDDS_INCLUDE_DIRS}/recordingservice"
+        ${CONNEXTDDS_INCLUDE_DIRS}
+        "${CONNEXTDDS_DIR}/include/recordingservice"
 )
 
 # Define the library that will provide the storage reader plugin
@@ -125,7 +127,8 @@ target_include_directories(
     PUBLIC
         "${CMAKE_CURRENT_BINARY_DIR}"
         "${CMAKE_BINARY_DIR}/generated"
-        "${CONNEXTDDS_INCLUDE_DIRS}/recordingservice"
+        ${CONNEXTDDS_INCLUDE_DIRS}
+        "${CONNEXTDDS_DIR}/include/recordingservice"
 )
 
 # Define the publisher application
--- a/examples/recording_service/pluggable_storage/c/CMakeLists.txt
+++ b/examples/recording_service/pluggable_storage/c/CMakeLists.txt
@@ -101,8 +101,6 @@ target_include_directories(
     PUBLIC
         "${CMAKE_CURRENT_BINARY_DIR}"
         "${CMAKE_BINARY_DIR}/generated"
-        ${CONNEXTDDS_INCLUDE_DIRS}
-        "${CONNEXTDDS_DIR}/include/recordingservice"
 )
 
 # Define the library that will provide the storage reader plugin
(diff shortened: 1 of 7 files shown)


examples/recording_service/pluggable_storage/c/FileStorageWriter.c
```diff
@@ -194,16 +194,26 @@ int FileStorageWriter_connect(void *storage_writer_data)
             (struct FileStorageWriter *) storage_writer_data;
     int64_t current_time = -1;
 
+#ifdef RTI_WIN32
+    fopen_s(&writer->file.file, writer->file.file_name, "w");
```
juanlu-rti, line 198: While I **hate** having to use different code when there is a standard library function, `fopen_s` is supposed to be better, more secure, less exploitable. In general, all the `_s` functions on Windows are better options than their standard counterparts.
Side question: in Osapi, for the functions that we define wrappers for, are we using this version of the functions on Windows?

examples/recording_service/pluggable_storage/cpp/FileStorageReader.cxx
```diff
@@ -15,6 +15,10 @@
 #include "dds/dds.hpp"
 #include <limits>
 
+#ifdef RTI_WIN32
+    #undef max
```
juanlu-rti, line 19: Wasn't there a `NOMINMAX` define that we could use or something similar?

examples/recording_service/pluggable_storage/cpp/FileStorageReader.cxx
```diff
@@ -220,11 +224,6 @@ bool FileStorageStreamReader::read_sample()
             throw std::runtime_error(
                     "Failed to read current data.msg field from file");
         }
-        std::getline(*data_file_, prefix);
```
juanlu-rti, line 223: What did this fragment even attempt to do?

juanlu-rti:
What were the missing symbols problems that you were seeing? Has that been fixed?

examples/recording_service/pluggable_storage/c/FileStorageWriter.c
Antoniolm: no, our file open function is a simple wrapper of fopen. Due to that, we will face the same warning issue.The strncat and strncpy function use their own implementation

examples/recording_service/pluggable_storage/cpp/FileStorageReader.cxx
Antoniolm: I think so, we can use either max and min or NOMINMAX. As I think that we will never use min on this scenario the  under max solution should be enough

examples/recording_service/pluggable_storage/cpp/FileStorageReader.cxx
Antoniolm: The problem was that we have two formats. The C format was storing the samples without an empty line between them and the C++ was storing the samples using an empty line between them. I faced some issues related to keeping this extra empty line on X64Win64VS2015. Due to that and for consistency purposes I prefer to use the C format.

Antoniolm:
Yes, the symbol problem was related to not including the macros.hpp header file as first RTI header file con StorageAdapter.hpp. This issue had been solved.

examples/recording_service/pluggable_storage/c/FileStorageWriter.c
```diff
@@ -35,8 +36,24 @@
 
 #define NANOSECS_PER_SEC 1000000000ll
 #define FileStorageWriter_FILE_NAME_MAX 1024
+#define FileStorageWriter_INFO_EXT ".info"
 #define FILENAME_PROPERTY_NAME "example.c_pluggable_storage.filename"
 
+int RTI_fopen(FILE **file, const char *file_name, const char *mode)
```
fgaranda, line 42: Maybe we should move this to a FileStorageUtils.c file that you share, so you don't need to copy the implementation.

examples/recording_service/pluggable_storage/cpp/FileStorageWriter.cxx
```diff
@@ -179,8 +179,8 @@ void PubDiscoveryFileStreamWriter::store(
 {
     using namespace dds::sub;
 
-    const int32_t count = sample_seq.size();
-    for (int32_t i = 0; i < count; ++i) {
+    const size_t count = sample_seq.size();
+    for (size_t i = 0; i < count; ++i) {
         const SampleInfo &sample_info = *(info_seq[i]);
         // we first first print the sample's metadata
         int64_t timestamp = (int64_t) sample_info->reception_timestamp().sec()
```
fgaranda, line 186: `static_cast<int64_t>`

examples/recording_service/pluggable_storage/cpp/FileStorageWriter.cxx
```diff
@@ -137,8 +137,8 @@ void FileStreamWriter::store(
     using namespace rti::core::xtypes;
     using namespace dds::sub;
 
-    const int32_t count = sample_seq.size();
-    for (int32_t i = 0; i < count; ++i) {
+    const size_t count = sample_seq.size();
+    for (size_t i = 0; i < count; ++i) {
         const SampleInfo &sample_info = *(info_seq[i]);
         // we first first print the sample's metadata
         int64_t timestamp = (int64_t) sample_info->reception_timestamp().sec()
```
fgaranda, line 144: `static_cast<>`

fgaranda (changes requested):

examples/recording_service/pluggable_storage/c/FileStorageUtils.c
```diff
@@ -0,0 +1,32 @@
+/*
+ * (c) 2019 Copyright, Real-Time Innovations, Inc.  All rights reserved.
```
fgaranda, line 2: 2021

examples/recording_service/pluggable_storage/c/FileStorageUtils.h
```diff
@@ -0,0 +1,17 @@
+/*
+ * (c) 2019 Copyright, Real-Time Innovations, Inc.  All rights reserved.
```
fgaranda, line 2: 2021

examples/recording_service/pluggable_storage/c/FileStorageReader.c
```diff
@@ -18,10 +18,12 @@
 #include <string.h>
 
 #include "ndds/ndds_c.h"
+#include "ndds/osapi/osapi_utility.h"
 #include "recordingservice/recordingservice_storagereader.h"
 #include "routingservice/routingservice_infrastructure.h"
```
fgaranda, line 23:
```suggestion

#include <ndds/ndds_c.h>
#include <osapi/osapi_utility.h>
#include <recordingservice_storagereader.h>
#include <routingservice/routingservice_infrastructure.h>
```

examples/recording_service/pluggable_storage/c/CMakeLists.txt
```diff
@@ -83,6 +83,7 @@ add_library(
     FileStorageWriter
         "${CMAKE_BINARY_DIR}/generated/HelloMsg.c"
         "${CMAKE_CURRENT_SOURCE_DIR}/FileStorageWriter.c"
+        "${CMAKE_CURRENT_SOURCE_DIR}/FileStorageUtils.c"
```
fgaranda, line 86: I'd add also the header files so that they trigger a recompilation if they change.

fgaranda (changes requested):

fgaranda (approved):

juanlu-rti (approved):

PR #519, opened 2022-07-07
Fixed Routing Service remote admin example

fgallegosalido:
### Summary
This pull request fixes the remote admin example in Routing Service and adds it to the CI.

### Details and comments
Renamed the example from `routing_service_remote_admin` to `remote_admin` for consistency. Adapted the example to work in the CI.

### Checks

-   [ ] I have read the [CONTRIBUTING](https://github.com/rticommunity/rticonnextdds-examples/blob/develop/CONTRIBUTING.md) document.


examples/routing_service/remote_admin/c++11/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@
 #  use or inability to use the software.
 #
 cmake_minimum_required(VERSION 3.11)
-project (RoutingServiceRemoteAdmin)
+project (RoutingServiceRemoteAdminC++11)
```
lulivi, line 13: Why did you change the project name? We use the same name in all the languages for a single example.

lulivi: But there will be no name collisions since we will not import one example into another.

lulivi (approved):

examples/routing_service/remote_admin/c++11/CMakeLists.txt
fgallegosalido: In any case, the examples related to the services are very inconsistent and may need a big rewrite. For now, we can leave them like that and in the future redesign all of them.

PR #520, opened 2022-07-08
Python Group Coherent Presentation Example

xFrenchy:
### Summary
Adding a group coherent presentation example using the Python API

### Details and comments
This example is a copy of the already existing group coherent presentation example (in c++) but translated with the new Python API binding. This example will randomly choose when to send samples and when to send a coherent set of samples.

Comments: Should I display how to set the presentation QoS for access_scope, coherent_access, and ordered_access? Also I'm not confident in my current README.md file, feels like I'm missing information. Especially for the "new to the python API, check out the getting started guide" area. Also in my publisher, in the get_patient_heart_rate function, the array in C++ is static and only created once. Not sure if my list in python will be created for every function call

### Checks


examples/connext_dds/group_coherent_presentation/py/GroupCoherentExample_publisher.py
```diff
@@ -0,0 +1,155 @@
...
+                # No alarm necessary, publish the patient's vitals as normal
+                heart_rate_writer.write(heart_rate_data)
+                temperature_writer.write(temperature_data)
+                print("Writing patient's vitals...")
+            time.sleep(2)
+        except KeyboardInterrupt:
+            break
+
+    print('preparing to shut down...')
+    alarm_writer.close()
```
alexcamposruiz, line 98: no need to call close on all these entities. You can just call participant.close() -- and even that's optional, but it ensures that it happens before leaving the function.

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExampleType.py
```diff
@@ -0,0 +1,52 @@
...
+@idl.enum
+class AlarmCode(IntEnum):
+    PATIENT_OK = auto()
+    PATIENT_REQUEST = auto()
+    SENSOR_OFFLINE = auto()
+    ABNORMAL_READING = auto()
+
+
+@idl.struct(
+    type_annotations={
```
alexcamposruiz, line 26:
```suggestion
    member_annotations={
```

Did you generate this with rtiddsgen?

alexcamposruiz:
Copy the IDL file into this folder.

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExample_publisher.py
```diff
@@ -0,0 +1,155 @@
...
+# fitness for any purpose. RTI is under no obligation to maintain or support
+# the Software.  RTI shall not be liable for any incidental or consequential
+# damages arising out of the use or inability to use the software.
+#
+import random
+import time
+import argparse
+
+import rti.connextdds as dds
+from GroupCoherentExampleType import AlarmCode, Alarm, HeartRate, Temperature
```
alexcamposruiz, line 16: When you generate the IDL file, you'll see that it's called GroupCoherentExample.py. The way we map modules from IDL to python is not quite the typical python way, so you'll have to do this:

```
from GroupCoherentExample import *

AlarmCode = GroupCoherentExample.AlarmCode
Alarm = GroupCoherentExample.Alarm
...
```

That's because GroupCoherentExample.py is itself module (derived from the IDL file name), and inside it defines a SimpleNamespace called GroupCoherentExample (the IDL module).

Maybe to keep the example simpler we should modify the IDL (only for Python) to remove the `module GroupCoherentExample` and then just this:

```
from GroupCoherentExample import AlarmCode, Alarm, HeartRate, Temperature
```

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExample_subscriber.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # Take all samples.  Samples are loaned to application, loan is
+    # returned when LoanedSamples destructor called or when you return
+    # the loan explicitly.
+    samples_read = 0
+    samples = reader.take()
+
+    for (data, info) in samples:
+        if info.valid:
+            print("Received: {}".format(data))
```
alexcamposruiz, line 27:
```suggestion
            print(f"Received: {data}")
```

Prefer `f`-strings for formatting

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExample_subscriber.py
```diff
@@ -0,0 +1,168 @@
...
+    temperature_topic = dds.Topic(participant, "Temperature", Temperature)
+
+    subscriber_qos = dds.QosProvider.default.subscriber_qos
+    subscriber = dds.Subscriber(participant, subscriber_qos)
+    reader_qos = dds.QosProvider.default.datareader_qos
+
+    # Create one Reader per Topic using the Subscriber and DataReader QoS
+    # specified in the xml file
+    alarm_reader = dds.DataReader(subscriber, alarm_topic, reader_qos)
+    heart_rate_reader = dds.DataReader(subscriber,
```
alexcamposruiz, line 49: Our coding style requires breaking the first parameter into a new line if you break any other parameter, so it would be:
```
 heart_rate_reader = dds.DataReader(
    subscriber,
    ...
```

alexcamposruiz (changes requested):
It looks pretty good! I'm providing a first round of feedback.

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExampleType.py
xFrenchy: I did not, I'll generate it from rtiddsgen so that all the type_annotiations are changed to member_annotations and instead of int, it will specify idl.int32

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExample_publisher.py
xFrenchy: Hmm, so should I generate the type file with rtiddsgen and then use the GroupCoherentExample module? Or should I  update my current type file (with modules not being used) and keep my current from file import TypeA, TypeB line?

xFrenchy: I've updated everything else. Let me know how I should proceed with this part. This is how I view it:

Use rtiddsgen to generate GroupCoherentExample.py -> update from import line -> Update publisher and subscriber to use correct scope -> Update README to explain how to generate GroupCoherentExample.py (although it will be included anyway)

or
Update GroupCoherentExampleType to use member_annotation instead of type_annotation and not use the module mapping but it keeps the example simpler

alexcamposruiz: Copy the IDL file to the Python directory, and remove the `module GroupCoherentExample` from the file. Then generate code and keep the import statement (it should work as-is).

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExample_publisher.py
```diff
@@ -0,0 +1,141 @@
...
+        verbosity = dds.Verbosity.WARNING
+    else:
+        verbosity = dds.Verbosity.STATUS_ALL
+
+    dds.Logger.instance.verbosity = verbosity
+
+    try:
+        run_publisher_application(args.domain, args.count)
+    except Exception as e:
+        print("Exception in run_publisher_application(): ", e)
```
alexcamposruiz, line 136: Use `f`-strings

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExample_subscriber.py
```diff
@@ -0,0 +1,163 @@
...
+        verbosity = dds.Verbosity.WARNING
+    else:
+        verbosity = dds.Verbosity.STATUS_ALL
+
+    dds.Logger.instance.verbosity = verbosity
+
+    try:
+        run_subscriber_application(args.domain, args.count)
+    except Exception as e:
+        print("Exception in run_subscriber_application(): ", e)
```
alexcamposruiz, line 158: Use `f`-strings

alexcamposruiz (changes requested):
Pending updated IDL file and a couple minor tasks.

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExample_subscriber.py
```diff
@@ -155,7 +155,7 @@ def main():
     try:
         run_subscriber_application(args.domain, args.count)
     except Exception as e:
-        print("Exception in run_subscriber_application(): ", e)
```
alexcamposruiz: Actually this wasn't bad or necessarily inconsistent. I got a bit overzealous :)

xFrenchy: Would you like me to revert it back to that print statement?

alexcamposruiz:
The destination branch must be develop, not master

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExample_subscriber.py
alexcamposruiz: No, this is fine too :)

examples/connext_dds/group_coherent_presentation/py/README.md
```diff
@@ -0,0 +1,23 @@
+# Example Code: Group Coherent Presentation
+
+If you haven't used the RTI Connext Python API before, first check the
+[Getting Started Guide](https://community.rti.com/static/documentation/connext-dds/6.0.1/api/connext_dds/api_python/index.html).
```
alexcamposruiz, line 4:
```
## Building the Example :wrench:
```

Section missing (just needs to say how to run rtiddsgen)

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExample_subscriber.py
```diff
@@ -0,0 +1,163 @@
+#
```
alexcamposruiz, line 1: I don't see `print_coherent_set_info` either. Review the C++ and C# examples. It looks like the logic is not the same in Python.

alexcamposruiz (changes requested):

examples/connext_dds/group_coherent_presentation/py/GroupCoherentExample_subscriber.py
alexcamposruiz: Also, run the example in C++ or C#, review the output and compare it to the ../README.md's Example Description and make sure you see a similar output and behavior in python.

alexcamposruiz (changes requested):
Example logic doesn't match C++/C#

alexcamposruiz:
You can't just change the destination branch. Now it's trying to merge your master branch fork into RTI's develop branch, adding 100s of unrelated commits. Try cherry-picking your commits into your develop branch and change the origin branch to develop.

PR #537, opened 2022-09-29
Python async pub example

xFrenchy:
### Summary
Python example displaying async publication

### Details and comments
!important! This example is not ready yet, creating initial pull request for review purposes. 
Here is a list of things to review:
-Idl file and class cannot be called "async" for python since it's a keyword. Naming the struct async and running rtiddsgen won't cause an error, however you won't be able to "from async import async". I have renamed it to async_type. Let me know if this should be approached differently
-When checking the history depth with Admin Console for the DataWriter, it shows a depth of 1 when it should be 12. This is observable when setting the QoS via XML and/or setting it programatically
-When creating a flow controller, I'm not sure what FlowController::FIXED_RATE_NAME is and what the python equivalent would be, for now I've created a flow controller without specifying a name

### Checks

### DIFF
--- a/examples/connext_dds/asynchronous_publication/py/async_publisher.py
+++ b/examples/connext_dds/asynchronous_publication/py/async_publisher.py
@@ -13,8 +13,6 @@ import time
 
 import rti.connextdds as dds
 
-# Note: cannot do 'from async' because async is a keyword since Python 3.5
-# therefore I have to update the name of async.idl to async_type.idl
 from async_type import async_type
 
 
@@ -70,9 +68,8 @@ def run_publisher_application(
     # You can wait until all written samples have been actually published
     # (note that this doesn't ensure that they have actually been received
     # if the sample required retransmission)
-    writer.wait_for_asynchronous_publishing
+    writer.wait_for_asynchronous_publishing()
     print("preparing to shut down...")
-    participant.close()
 
 
 def main():
--- a/examples/connext_dds/asynchronous_publication/py/async_publisher.py
+++ b/examples/connext_dds/asynchronous_publication/py/async_publisher.py
@@ -1,0 +1,127 @@
+#
+# (c) 2022 Copyright, Real-Time Innovations, Inc.  All rights reserved.
+#
+# RTI grants Licensee a license to use, modify, compile, and create derivative
+# works of the Software solely for use with RTI products.  The Software is
+# provided "as is", with no warranty of any type, including any warranty for
+# fitness for any purpose. RTI is under no obligation to maintain or support
+# the Software.  RTI shall not be liable for any incidental or consequential
+# damages arising out of the use or inability to use the software.
+#
+import argparse
+import time
+
+# Note, I have not run this program yet. Need to pull on my local laptop so that I can download the wheel
+import rti.connextdds as dds
+# Note: cannot do 'from async' because async is a keyword since Python 3.5 
+# therefore I have to update the name of async.idl to async_type.idl
+from async_type import async_type
+
+
+def run_publisher_application(domain_id: int, sample_count: int):
+    # Start communicating in a domain, usually one participant per application
+    participant = dds.DomainParticipant(domain_id)
+
+    # Create a Topic
+    topic = dds.Topic(participant, "Example async", async_type)
+
+    # Retrieve the default DataWriter QoS, from USER_QOS_PROFILES.xml
+    writer_qos = dds.QosProvider.default.datawriter_qos
+
+    # Note: Update this to Python
+    """ # If you want to change the DataWriter's QoS programmatically rather than
+    # using the XML file, uncomment the following lines.
+
+    # In this case, we set the publish mode QoS to asynchronous publish mode,
+    # which enables asynchronous publishing. We also set the flow controller
+    # to be the fixed rate flow controller, and we increase the history depth.
+    writer_qos << Reliability::Reliable(Duration(60));
+
+    DataWriterProtocol writer_protocol_qos;
+    RtpsReliableWriterProtocol rtps_writer_protocol;
+    rtps_writer_protocol.min_send_window_size(50);
+    rtps_writer_protocol.max_send_window_size(50);
+    writer_protocol_qos.rtps_reliable_writer(rtps_writer_protocol);
+    writer_qos << writer_protocol_qos;
+
+    # Since samples are only being sent once per second, DataWriter will need
+    # to keep them on queue.  History defaults to only keeping the last
+    # sample enqueued, so we increase that here.
+    writer_qos << History::KeepLast(12);
+
+    # Set flowcontroller for DataWriter
+    writer_qos << PublishMode::Asynchronous(FlowController::FIXED_RATE_NAME); """
+
+    # Create a DataWriter with the QoS in our profile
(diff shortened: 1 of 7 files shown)


examples/connext_dds/asynchronous_publication/py/async_publisher.py
```diff
@@ -0,0 +1,126 @@
...
+        try:
+            # Modify the data to be sent here
+            sample.x = count
+            print("Writing async_type, count {}".format(count))
+            writer.write(sample)
+            time.sleep(1)
+        except KeyboardInterrupt:
+            break
+    print("preparing to shut down...")
+    writer.close()
```
alexcamposruiz, line 70: explicit calls to close are not necessary

examples/connext_dds/asynchronous_publication/py/async_publisher.py
```diff
@@ -0,0 +1,126 @@
...
+    writer = dds.DataWriter(participant.implicit_publisher, topic, writer_qos)
+
+    # Create data sample for writing
+    sample = async_type(x=42)
+
+    for count in range(sample_count):
+        try:
+            # Modify the data to be sent here
+            sample.x = count
+            print("Writing async_type, count {}".format(count))
```
alexcamposruiz, line 64: use f-strings for formatting

examples/connext_dds/asynchronous_publication/py/async_publisher.py
```diff
@@ -0,0 +1,126 @@
+#
```
alexcamposruiz, line 1: I think you should base the example on the C# one, which is more complete:

- it provides an option to use XML or programmatic Qos
- It uses WaitForAsyncPublication

alexcamposruiz (changes requested):

alexcamposruiz:
You should now target master instead of develop, since this example will work with Connext 7.0.0

lulivi:
Hi @xFrenchy ! Can you apply the `black` formatting to this example?

You can run:

```bash
pip install black==22.3.0
black -l 79 --preview <file or directory>
```

xFrenchy:
I applied black formatting (I think). I left a few notes in the publisher and subscriber files of things that I've found that I think are problematic. Right now the most problematic is the subscriber throwing an exception when shutting down the application. Otherwise I think it follows the C# example better than the previous attempt. I added an additional argument to use the QoS file or not. I didn't see this implemented in the C# example though. I see the variable, but I never see an argument or a way to toggle that variable in the example.

alexcamposruiz (changes requested):
I'm leaving some feedback. I still have to complete the review and test or confirm the issues you mention throughout the code.

examples/connext_dds/asynchronous_publication/py/async_subscriber.py
```diff
@@ -0,0 +1,121 @@
...
+        print("async subscriber sleeping")
+        rti.asyncio.run(process_data(reader))
+        # await process_data(reader)
+    except KeyboardInterrupt:
+        pass
+
+    print("preparing to shut down...")
+    # rti.asyncio.close(process_data(reader))
+    # Note: Exception thrown here. I assume it's because the coroutine is still running
+    participant.close()
```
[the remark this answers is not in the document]
xFrenchy: Yup, my bad. I don't know why I keep forgetting that. I removed both.

examples/connext_dds/asynchronous_publication/py/async_type.idl
```diff
@@ -0,0 +1,15 @@
...
+ * RTI grants Licensee a license to use, modify, compile, and create derivative
+ * works of the Software.  Licensee has the right to distribute object form
+ * only for use with RTI products.  The Software is provided "as is", with no
+ * warranty of any type, including any warranty for fitness for any purpose.
+ * RTI is under no obligation to maintain or support the Software.  RTI shall
+ * not be liable for any incidental or consequential damages arising out of the
+ * use or inability to use the software.
+ */
+
+ struct async_type {
```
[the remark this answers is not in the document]
xFrenchy: Should I then remove the IDL from the repo?

examples/connext_dds/asynchronous_publication/py/async_type.py
```diff
@@ -0,0 +1,18 @@
+# WARNING: THIS FILE IS AUTO-GENERATED. DO NOT MODIFY.
```
alexcamposruiz, line 1: Replace this with the (c) header. It's actually OK to modify this file.

examples/connext_dds/asynchronous_publication/py/async_publisher.py
```diff
@@ -0,0 +1,134 @@
...
+    if use_xml_qos:
+        # Retrieve the default DataWriter QoS, from USER_QOS_PROFILES.xml
+        writer_qos = dds.QosProvider.default.datawriter_qos
+    else:
+        # Configure the DataWriterQos in code, to the same effect as the XML file.
+        writer_qos = participant.default_datawriter_qos
+        writer_qos.reliability = dds.Reliability.reliable(
+            dds.Duration.from_seconds(60)
+        )
+        # Note: Admin Console is reporting history depth of 1 still regardless of changes in the QoS file or programatically
```
alexcamposruiz, line 37: I verified that the depth is 12, but the propagated BuiltinTopicData says it's 1, even in a C++ app. 

Did you see any issue in the example related to incorrect History Qos? If not, you can delete this note.

examples/connext_dds/asynchronous_publication/py/async_publisher.py
```diff
@@ -0,0 +1,134 @@
...
+        writer_qos.history.kind = dds.HistoryKind.KEEP_LAST
+        writer_qos.history.depth = 12
+        writer_qos.data_writer_protocol.rtps_reliable_writer.min_send_window_size = (
+            50
+        )
+        writer_qos.data_writer_protocol.rtps_reliable_writer.max_send_window_size = (
+            50
+        )
+
+        # Note: Without specifying a name it defaults to DDS_DEFAULT_FLOW_CONTROLLER_NAME. Do I need to specify the name?
```
alexcamposruiz, line 48: Yes, you have to specify the name. We need to add a constant with the fixed rate name, but for now use this string.

examples/connext_dds/asynchronous_publication/py/async_publisher.py
```diff
@@ -0,0 +1,134 @@
...
+        3: dds.Verbosity.STATUS_ALL,
+    }
+
+    # Sets Connext verbosity to help debugging
+    verbosity = verbosity_levels.get(args.verbosity, dds.Verbosity.EXCEPTION)
+
+    dds.Logger.instance.verbosity = verbosity
+
+    try:
+        # Note: C# example doesn't seem to allow using the use_xml_qos option?
```
alexcamposruiz, line 126: It looks like this example didn't include that option. It's not a big deal.

alexcamposruiz (changes requested):

examples/connext_dds/asynchronous_publication/py/async_publisher.py
xFrenchy: I didn't see any issue, it was more of an observation I found when trying to confirm that the QoS changes I was making were correctly applying.

alexcamposruiz (approved):

PR #640, opened 2023-11-24
Adding dynamic data enum example

angelrti:
### Summary
This PR adds an example that shows how to use DynamicData Enums and get their member ordinal values by name (string).

### Checks

-   [X] I have added a new C/C++ example and updated `examples/connext_dds/CMakeList.txt` accordingly.

### DIFF
--- a/examples/connext_dds/dynamic_data_get_enum_value/c++11/dynamic_data_enum_example.cxx
+++ b/examples/connext_dds/dynamic_data_get_enum_value/c++11/dynamic_data_enum_example.cxx
@@ -41,7 +41,6 @@ int32_t get_enum_ordinal_value_by_name(
         DynamicType struct_dynamic_type,
         std::string enum_element)
 {
-
     // Get value using scenario 1 of `print_enum_ordinal_value_by_name`
     const StructType &dynamic_struct =
             static_cast<const StructType &>(struct_dynamic_type);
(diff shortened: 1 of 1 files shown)


examples/connext_dds/dynamic_data_get_enum_value/c++11/dynamic_data_enum_example.cxx
```diff
@@ -0,0 +1,165 @@
...
+using namespace dds::core::xtypes;
+
+EnumType create_enum_type()
+{
+    return EnumType(
+            "EngineState",
+            { EnumMember("off", 0), EnumMember("on", 42) });
+}
+
+DynamicType create_struct_dynamic_type()
```
alexcamposruiz, line 26: It can be simplified:

```
{
    return StructType("EnumStruct", { {"engine_state_element", create_enum_type() } });
}
```

alexcamposruiz, line 26:
```suggestion
StructType create_struct_dynamic_type()
```

alexcamposruiz:
This looks like a really specific and small example. I'm not sure it makes sense to create a brand new example instead of adding the relevant code to the documentation or incorporating it into another example

alexcamposruiz (changes requested):

PR #641, opened 2023-12-12
Connext Secure Examples

akkshaj-rti:
### Summary

Added examples for Connext Secure

### Details and comments

Added 4 examples showing new features in Connext Secure 7.2.0 showing lightweight security, certificate revocation, and whitelisting.

### Checks

-   [x] I have added a new C/C++ example and updated `examples/connext_dds/CMakeList.txt` accordingly.


examples/CMakeLists.txt
```diff
@@ -34,20 +34,33 @@ option(CONNEXTDDS_BUILD_WEB_INTEGRATION_SERVICE_EXAMPLES
     OFF
 )
 
+option(CONNEXTDDS_BUILD_CONNEXT_SECURE_EXAMPLES
+    "Build Connext Secure examples"
+    OFF
```
alexcamposruiz, line 39: @lulivi I'm not sure this needs to be disabled or not. Can you advise?

colmenero: I guess it depends on whether we can build the new examples with the current GitHub pipeline. I'd assume it would not be too complicated to do so, with the exception of running the `setup.py` python script. This is a different step that in other examples, so it may not be straight-forward. Let's wait for @lulivi's opinion on it.

examples/connext_secure/whitelist/security/identities/ParticipantC/ParticipantC.pem
colmenero, line 1: @akkshaj-rti The idea of having the python `setup.py` script was (among other things) to prevent polluting the git history with security artifacts.

The instructions already tell the user how to generate them, so we only need here the source files that `setup.py` needs (like `ParticipantC.cnf`)

examples/CMakeLists.txt
lulivi: We currently don't support the `setup.py` Python script as @colmenero mentioned. So I would leave this OFF for now. We can open an issue in order to configure a step of running setup before the examples.
Also I would suggest a different name for the script since it is the common name for any Python package entry point (https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#setup-py) and could cause confusion and possible problems with some tools.

examples/connext_secure/whitelist/security/identities/ParticipantC/ParticipantC.pem
akkshaj-rti: Right, I think I missed the .gitignore when copying them over. Removed.

examples/CMakeLists.txt
akkshaj-rti: Renaming to `setup_security.py`

lulivi (approved):

examples/connext_secure/lightweight/security/ca/Ca.cnf
```diff
@@ -0,0 +1,86 @@
...
+certificate     = $dir/CaCert.pem
+private_key     = $dir/CaKey.pem
+crl_dir			= $dir/crl
+
+new_certs_dir   = ./temporary_files
+database        = $dir/database/CaIndex
+crlnumber       = $dir/crl/crlNumber
+
+# Due to cmake code, we need to use certs/serial
+# serial          = cert/temporary_files/serial
```
colmenero, line 23:
```suggestion
```
Remove if not used. Same for line above, which doesn't seem to apply here

examples/connext_secure/certificate_revocation_list/c++11/README.md
```diff
@@ -0,0 +1,107 @@
+# Example Code: Using a Certificate Revocation List
+
+## Building the Example
+
+Remember to set your environment variables with the script in your Connext
+installation directory before building.
+
+```sh
+cd ../../c++11/
```
colmenero, line 9:
```suggestion
cd c++11/
```

examples/connext_secure/certificate_revocation_list/c++11/README.md
```diff
@@ -0,0 +1,111 @@
...
+To restart communication, call the script again to give the participants new
+certificates. You can do this without restarting the applications.
+
+First, delete the contents of `build/security/ca/database/CaIndex`.
+Alternatively, you can delete the file and create a new empty one.
+Then, run the following to give the participants new certificates.
+
+```sh
+cd build/security
+python3 setup.py
```
colmenero, line 103:
```suggestion
python3 setup_security.py
```

colmenero (approved):

alexcamposruiz (approved):

PR #642, opened 2023-12-15
Add dynamic_data_skip_serialization example

alexcamposruiz:
### Summary

An example for the upcoming DynamicData feature that allows skipping the serialization/deserialization of data.

### Details and comments

* Added README with example description
* Added C++11 example
* Added Python example

### Checks

* Tested all modes in both languages
* Tested interoperability between the examples in both languages

### DIFF
--- a/examples/connext_dds/dynamic_data_skip_serialization/README.md
+++ b/examples/connext_dds/dynamic_data_skip_serialization/README.md
@@ -17,7 +17,7 @@ where a custom Recording application may be useful.
 
 This example implements a simple recording application that uses the DynamicData
 API to receive data in CDR format and directly record it in a file. The example
-also provides an replay option that reads the data buffers from the file
+also provides a replay option that reads the data buffers from the file
 and publishes them back. For convenience, an option to publish a few samples
 to test the record and replay functionality is also provided.
 
--- a/examples/connext_dds/dynamic_data_skip_serialization/README.md
+++ b/examples/connext_dds/dynamic_data_skip_serialization/README.md
@@ -1,0 +1,29 @@
+# Example Code: Skip Data Serialization for Data Recording
+
+## Concept
+
+There are scenarios where the need to deserialize or inspect data is not necessary.
+A common example is recording data. In this case, the application can simply
+record the binary data as it is received, and then replay it later. This provides
+a significant performance improvement.
+
+The DynamicData API provides a mode that allows sending or receiving data in its
+CDR format, without serializing or deserializing it.
+
+Note that while Connext provides a Recording Service, there may be situations
+where a custom Recording application may be useful.
+
+## Example Description
+
+This example implements a simple recording application that uses the DynamicData
+API to receive data in CDR format and directly record it in a file. The example
+also provides an replay option that reads the data buffers from the file
+and publishes them back.
+
+The key parts of the example are implemented in the ``record()`` and
+``replay()`` functions in the example source code for each language.
+
+The example is very simple and uses a single type and topic, but it could be
+extended to use discovered types (see the
+[built-in topics example](../builtin_topics/)) to implement a more
+general-purpose recording application.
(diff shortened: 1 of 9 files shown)


examples/connext_dds/dynamic_data_skip_serialization/README.md
```diff
@@ -0,0 +1,30 @@
+# Example Code: Skip Data Serialization for Data Recording
```
mcmanuea-rti, line 1: Since the field is called skip_deserialization I would suggestion renaming the example to dynamic_data_skip_deserialization example
and change this line to say 'Skip Data Deserialization for Data Recording' since for recording we're not skipping serialization, we're skipping deserialization

examples/connext_dds/dynamic_data_skip_serialization/README.md
```diff
@@ -0,0 +1,30 @@
...
+CDR format, without serializing or deserializing it.
+
+Note that while Connext provides a Recording Service, there may be situations
+where a custom Recording application may be useful.
+
+## Example Description
+
+This example implements a simple recording application that uses the DynamicData
+API to receive data in CDR format and directly record it in a file. The example
+also provides an replay option that reads the data buffers from the file
```
mcmanuea-rti, line 20:
```suggestion
also provides a replay option that reads the data buffers from the file
```

examples/connext_dds/dynamic_data_skip_serialization/c++11/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
+#
+# (c) 2018 Copyright, Real-Time Innovations, Inc.  All rights reserved.
```
mcmanuea-rti, line 2: copyright should be updated

examples/connext_dds/dynamic_data_skip_serialization/c++11/README.md
```diff
@@ -0,0 +1,69 @@
+# Example Code: Skip Data Serialization for Data Recording
```
mcmanuea-rti, line 1:
```suggestion
# Example Code: Skip Data Deserialization for Data Recording
```

examples/connext_dds/dynamic_data_skip_serialization/c++11/recorder.cxx
```diff
@@ -0,0 +1,194 @@
...
+}
+
+void replay(const std::string &file_name, int domain_id)
+{
+    using dds::core::xtypes::DynamicData;
+
+    dds::domain::DomainParticipant participant(domain_id);
+
+    // For the replay application we don't need to register the type with any
+    // particular property
```
mcmanuea-rti, line 112:
```suggestion
    // particular property because DynamicData DataWriters are always prepared to write serialized CDR buffers directly
```

mcmanuea-rti (changes requested):

examples/connext_dds/dynamic_data_skip_serialization/README.md
alexcamposruiz: I use "serialization" to refer to the general process to transmit data on the wire, rather than the specific direction. I'm trying to avoid saying "serialization and deserialization".

The example skips both directions, so while the most "correct" and specific title would be "Skip Data Deserialization for Data Recording and Data Serialization for Data Replay", I think "Skip Data Serialization for Data Recording" is a way to be more general and include both.

examples/connext_dds/dynamic_data_skip_serialization/c++11/README.md
alexcamposruiz: see my other comment as to why I'd prefer to keep "Serialization"

examples/connext_dds/dynamic_data_skip_serialization/README.md
mcmanuea-rti: I figured that was the reason. I'm okay with it. I was thinking if a customer was looking at the field 'skip_deserialization' and looking for an example, the name doesn't match exactly, but hopefully they would be able to make that leap...

mcmanuea-rti (approved):

hman523 (approved):

PR #643, opened 2024-01-05
PR for Routing Service example for future blogpost

franporcel:
### Checks


lulivi:
Hi @franporcel ! Can you add new lines to the end of the scripts that are missing and fix the markdown issues (https://github.com/rticommunity/rticonnextdds-examples/actions/runs/7427180207/job/20212388518?pr=643)?

franporcel:
Thanks @lulivi ! I fixed all the issues (I think)

examples/routing_service/shmem_udp_gateway/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+For your convenience, there are 2 scripts to set up environment variables:
+
+- Linux: *variables.sh*
+- Windows: *variables.bat*
+
+These are the variables they contain, which you should modify according to your
+system:
+
+-   **NDDSHOME**: path to the installation of RTI Connext Professional.
```
alexcamposruiz, line 20: Instead of using NDDSHOME in the instructions, ask the user to run the rtisetenv script. See other examples, for ex:

To build this example, first set all the environment variables as follows:

On *Linux* and *macOS* systems run:

```sh
source <install_dir>/recource/scripts/rtisetenv_<target>.<bash|tcsh|zsh>
```

On *Windows* systems run:

```sh
<install_dir>\resource\scripts\rtisetenv_<target>.bat
```

examples/routing_service/shmem_udp_gateway/README.md
```diff
@@ -0,0 +1,172 @@
+# SHMEM <--> UDP Gateway
+
+This repository contains the necessary files to run the RTI Routing Service
+example from the "How does Connext help you meet the constraints of the
+hospital's IT department?" blogpost. Using RTI DDS Ping as a pub/sub
+application and Routing Service you will be able to create a gateway between
+a Shared Memory domain and a UDP domain that uses only 3 ports. There is also
+a second configuration option using RTI Cloud Discovery Service to use Unicast.
```
alexcamposruiz, line 8: Without fully duplicating the blogpost, can you add a short motivation for why you would want a gateway here?

alexcamposruiz (changes requested):
I've done a high-level review. We still need someone to review the XML files and Routing Service / CDS commands

examples/routing_service/shmem_udp_gateway/README.md
franporcel: @alexcamposruiz good point. I thought about it, but for this example the only thing that's needed is the *rtiroutingservice* binary, there's no compilation necessary so NDDSHOME is enough. But if you still want me to add it, I will. Just let me know

franporcel:
Thanks @alexcamposruiz !!

examples/routing_service/shmem_udp_gateway/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Multicast example
+
+You will need 3 terminals to run this example.
+
+1.  On terminal 1, source the variables script and run an RTI DDS Ping
+publisher on SHMEM acting as a local publisher:
+
+    ```bash
+    source variables.sh
```
fgaranda, line 63: I'd remove the dependency on this file altogether. You could do the following:

```bash
<install_dir>/bin/rtiddsping -pub -domain 0 \
    -qosFile <example_dir>/qos.xml \
    -qosProfile "example_library::smem_profile"

<install_dir>/bin/rtiddsping -sub -domain 1 \
    -qosFile <example_dir>/qos.xml \
    -qosProfile "example_library::multicast"
```
You could simply point to the configuration tweaks in the XML if you want to change APPs and CDS port and IP address.

As for Routing Service, you can provide the value of environment variables via command line argument as follows, which I think it's pretty neat  (this is what we do in our Docker images):

```bash
<install_dir>/bin/rtiroutingservice  \
    -cfgFile <example_dir>/RS_config_multicast.xml \
    -cfgName gateway_SHMEM_and_UDP \
    -DSHMEM_DOMAIN=0 =DUDP_DOMAIN=1
```

Same with Cloud Discovery Service:

````bash
<install_dir>/bin/rticlouddiscoveryservice -cfgFile CDS_config.xml -cfgName cds_all_domains_udpv4 \
    -DSHMEM_DOMAIN=0 =DUDP_DOMAIN=1
````

File: examples/routing_service/shmem_udp_gateway/qos.xml [hunk 4]
```diff
@@ -0,0 +1,54 @@
...
+to use the software. -->
+
+<dds xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://community.rti.com/schema/7.2.0/rti_dds_qos_profiles.xsd">
+    <qos_library name="example_library">
+
+        <qos_profile name="shmem_profile" base_name="BuiltinQosLib::Generic.Common">
+            <domain_participant_qos>
+                <discovery>
+                    <initial_peers>
+                        <element>$(APPS)@builtin.shmem://</element>
```
fgaranda, line 20: You should probably set a default for each environment variable as follows:

```xml
<configuration_variables>
    <value>
        <element>
            <name>APPS</name>
            <value>5</value>
        </element>
        <element>
            <name>CDS_IP_ADDRESS</name>
            <value>127.0.0.1</value>
        </element>
        <element>
            <name>CDS_PORT</name>
            <value>8000</value>
        </element>
    </value>
</configuration_variables>
```

examples/routing_service/shmem_udp_gateway/variables.bat
```diff
@@ -0,0 +1,17 @@
+REM Multicast example
+
+set NDDSHOME="C:/Program Files/rti_connext_dds-7.2.0"
```
fgaranda, line 3: Like I mentioned before, this file would no longer be necessary. Also, this assumes you're using that installation directory and version 7.2.0, which is not always going to be the case.

fgaranda (changes requested):

examples/routing_service/shmem_udp_gateway/README.md
franporcel: @fgaranda I don't fully agree with your first recommendation. Both ways definitely work, but using environment variables provides flexibility since the customer doesn't need to hardcode magic numbers. At the end of the day, we need to show the customer how to use best practices, and using environment variables is one of them. Specially since every host in the network in this scenario will need different values for those environment variables.

@fgaranda wrt your second recommendation, I agree that using -D is a cool feature, but using "centralized" environment variables in a file gives the user full control. They can use the same method for both their applications (XML QoS) and their Infrastructure Services (Routing Service and CDS). Whereas hardcoding values on XML and using -D for IS are 2 different methods to do the same thing.

examples/routing_service/shmem_udp_gateway/qos.xml
franporcel: @fgaranda I wouldn't be comfortable suggesting this to a customer. This can lead to inconsistencies if a developer "blindly" trusts these default variables without setting them outside. It is a general services recommendation to have the application complain and fail if variables are not set. Using the <configration_variables> tag goes against that recommendation, since it'll allow the application to just work and not complain

examples/routing_service/shmem_udp_gateway/variables.bat
franporcel: @fgaranda I could change the README file and the environment variables script to say something like: "Feel free to change the values of the environment variables and adapt them to your system." Would that be okay with you?

alexcamposruiz (approved):

fgaranda (approved):

fgaranda (approved):

fgaranda (approved):

PR #661, opened 2024-02-21
Add automerge.yml

ManuelJNunez:
### Summary

Add automerge.yml, this workflow file should be placed on the default branch (doc [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule)). We shouldn't merge this file into other branches. Let's be careful, please.

### Checks

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -1,0 +1,30 @@
+name: Automerge
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0/12 * * *'
+
+env:
+  REPO: https://rticommunity:${{secrets.GITHUB_TOKEN}}@github.com/rticommunity/rticonnextdds-examples.git
+
+  DESTINATION_BRANCH: develop
+
+  MASTER_REPO: https://github.com/rticommunity/rticonnextdds-examples.git
+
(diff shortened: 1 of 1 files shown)


.github/workflows/automerge.yml
```diff
@@ -0,0 +1,30 @@
...
+  merge:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Merge with master
+      run: |
+        git clone ${{env.REPO}} -b ${{env.DESTINATION_BRANCH}} tmp
+        cd tmp
+        git config user.name "Automerge Bot"
+        git config user.email "bot@example.com"
```
ManuelJNunez, line 27: I don't know what email to use... maybe automerge@rti.com?

.github/workflows/automerge.yml
```diff
@@ -0,0 +1,30 @@
+name: Automerge
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0/12 * * *'
```
ManuelJNunez, line 6: This workflow will be executed every 12 hours.

.github/workflows/automerge.yml
lulivi: @alexcamposruiz @fgaranda

PR #697, opened 2024-07-19
ROUTING-1234: having the Service as Lib example in Recorder show the …

juanlu-rti:
### Summary

Modify some CMake files in recording service examples so that they use the newly shipped library (7.4.0) `rtiserviceadmincpp`. Also, modified the example `service_as_lib` so that it showcases the use of the `execute_command` method to shut down the service.

### Checks


examples/recording_service/service_as_lib/c++11/ServiceAsLibExample.cxx
```diff
@@ -109,7 +109,24 @@ int main(int argc, char *argv[])
         embedded_service.start();
         // Wait for 'running_seconds' seconds
         std::this_thread::sleep_for(std::chrono::seconds(running_seconds));
-        embedded_service.stop();
+        // embedded_service.stop();
```
fgaranda, line 112:
```suggestion
```

Should it be removed?

fgaranda (dismissed):

juanlu-rti:
I have retargeted to develop like we mentioned

juanlu-rti:
I don't understand, because I applied the feedback provided by @fgaranda and resolved the comment, yet it says that I have changes requested.
@lulivi , do you think you can review this this week?

juanlu-rti:
@fgaranda is on vacation, I'm going to take him out of the review as he provided feedback and has been applied.

lulivi (approved):

PR #698, opened 2024-07-22
Update RPC example to use new features

alexcamposruiz:
### Summary

Update RPC example to use new features.

### Details and comments

- Service discovery
- Built-in RPC QoS profile

### DIFF
--- a/examples/connext_dds/remote_procedure_call/c++11/USER_QOS_PROFILES.xml
+++ b/examples/connext_dds/remote_procedure_call/c++11/USER_QOS_PROFILES.xml
@@ -1,9 +1,10 @@
 <?xml version="1.0"?>
 
 <dds xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:noNamespaceSchemaLocation="http://community.rti.com/schema/7.0.0/rti_dds_qos_profiles.xsd">
+    xsi:noNamespaceSchemaLocation="http://community.rti.com/schema/current/rti_dds_profiles.xsd">
 
     <qos_library name="RequestReplyExampleProfiles">
+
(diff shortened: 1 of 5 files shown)


examples/connext_dds/remote_procedure_call/c++11/USER_QOS_PROFILES.xml
```diff
@@ -4,147 +4,33 @@
...
-        
-            <!-- QoS for the data writer that sends replies -->
-            <datawriter_qos>
-                <durability>
-                    <kind>VOLATILE_DURABILITY_QOS</kind>
-                </durability>
-            </datawriter_qos>
+        <qos_profile name="ReplierExampleProfile" base_name="BuiltinQosLib::Pattern.RPC">
 
-            <!-- QoS for the data reader that receives requests -->
```
jmuros-rti, line 142: Why are you removing the text but keep the comment? 
I think that we should keep a comment with the meaning or remove the whole comment.

examples/connext_dds/remote_procedure_call/py/README.md
```diff
@@ -1,7 +1,7 @@
 # Example Code: Remote Procedure Calls
 
 :warning: **Note**: For an overview of this feature, see the
```
jmuros-rti, line 3: Should we mention that the Python RPC is experimental?

jmuros-rti (changes requested):
See the two questions

examples/connext_dds/remote_procedure_call/py/README.md
alexcamposruiz: We already say that in the documentation. I don't think we need to say it here.

jmuros-rti (approved):

PR #703, opened 2024-08-26
EVALG-77 follow up: Change design to use Last-Value Cache and Read

Alxe:
### Summary

This PR changes the example to follow a Last-Value Cache pattern on the subscriber. This removes the need for asynchronous code and focuses on polling.

Additionally, a few other changes have been made as left-over TODO from previous PR, #695.

### Checks

### DIFF
--- a/tutorials/application_design/py/subscriber.py
+++ b/tutorials/application_design/py/subscriber.py
@@ -58,14 +58,15 @@ class SubscriberDashboard:
 
         dashboard_condition.set_handler(dashboard_handler)
 
-
         display_handler_sentinel = threading.Event()
         display_handler_condition = threading.Condition()
+
         def display_handler():
             while not display_handler_sentinel.is_set():
                 with display_handler_condition:
                     display_handler_condition.wait(5)
                     dashboard_condition.trigger_value = True
+
         display_thread = threading.Thread(target=display_handler)
 
         waitset = dds.WaitSet()
--- a/tutorials/application_design/py/subscriber.py
+++ b/tutorials/application_design/py/subscriber.py
@@ -86,7 +86,9 @@ class SubscriberDashboard:
 
     def _create_new_position_string(self, condition):
         string = ""
-        for sample, info in self._transit_reader.select().condition(condition).read():
+        for sample, info in (
+            self._transit_reader.select().condition(condition).read()
+        ):
             if not info.valid:
                 continue
 
@@ -94,9 +96,7 @@ class SubscriberDashboard:
             if sample.current_route and len(sample.current_route):
                 string += f" is en route to {sample.current_route[-1]} from {sample.current_position}"
             else:
-                string += (
-                    f" has arrived at its destination in {sample.current_position}"
-                )
+                string += f" has arrived at its destination in {sample.current_position}"
             string += "\n"
         return string
 
@@ -116,7 +116,9 @@ class SubscriberDashboard:
             f"  Last known fuel level: {data.fuel_history[-1]}\n"
             for data in online_vehicles
         )
-        offline_str = "\n".join(f"Vehicle {data.vin}" for data in offline_vehicles)
+        offline_str = "\n".join(
+            f"Vehicle {data.vin}" for data in offline_vehicles
+        )
 
         return "\n".join(
             [
--- a/tutorials/application_design/py/subscriber.py
+++ b/tutorials/application_design/py/subscriber.py
@@ -10,7 +10,7 @@
 #
 
 import dataclasses
-import time
+import threading
 import typing
 from datetime import datetime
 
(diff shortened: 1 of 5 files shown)


tutorials/application_design/py/subscriber.py
```diff
@@ -45,71 +38,110 @@ def __init__(
...
+                f"[[ DASHBOARD: {datetime.now()} ]]",
+                f"Online vehicles: {len(online_vehicles)}",
+                online_str,
+                f"Offline vehicles: {len(offline_vehicles)}",
+                offline_str,
+            ]
+        )
+
+    def _dashboard_data(self):
+        data: typing.Dict[dds.InstanceHandle, DashboardItem] = {}
```
alexcamposruiz, line 133: If you use typing for a variable here, you should also use it for the parameters and return types.

Also, prefer `data: Dict[..., ...]`, and `from typing import Dict, ...`.

tutorials/application_design/py/subscriber.py
```diff
@@ -45,71 +38,110 @@ def __init__(
...
+                waitset.dispatch()
+                dashboard_condition.trigger_value = False
+        except KeyboardInterrupt:
+            with display_handler_condition:
+                display_handler_sentinel.set()
+                display_handler_condition.notify_all()
+
+        display_thread.join()
+
+    def _new_position_string(self, condition):
```
alexcamposruiz, line 88: Would it make more sense to have the conditions as instance variables (in `self`) instead of as arguments? Like the readers, they can be set up in the `__init__` method and won't change.

tutorials/application_design/py/subscriber.py
```diff
@@ -45,71 +38,110 @@ def __init__(
...
-            print()
-            await asyncio.sleep(0.5)
-
-    async def _metrics_app(self):
-        async for sample, info in self._metrics_reader.take_async():
-            if info.instance_handle not in self._dashboard_data:
+            string += "\n"
+        return string
+
+    def _dashboard_string(self):
```
alexcamposruiz, line 104: Use verbs for methods e.g.: `_create_dashboard_string` or `_render_dashboard`

tutorials/application_design/py/subscriber.py
```diff
@@ -45,71 +38,110 @@ def __init__(
     ):
```
alexcamposruiz, line 38: why is dds.DataReader between quotes?

Should be:

```
    def __init__(
        self,
        metrics_reader: dds.DataReader,
        transit_reader: dds.DataReader,
    ) -> None:
```

tutorials/application_design/py/subscriber.py
```diff
@@ -45,71 +38,110 @@ def __init__(
...
         self._metrics_reader = metrics_reader
         self._transit_reader = transit_reader
-        self._dashboard_data: typing.Dict[
-            dds.InstanceHandle, DashboardItem
-        ] = dict()
 
     def __repr__(self):
-        return f"Dashboard({self._metrics_reader=}, {self._transit_reader=}, {self._dashboard_data=})"
+        return f"Dashboard({self._metrics_reader=}, {self._transit_reader=})"
```
alexcamposruiz, line 43: Does the string representation of the DataReader produce anything useful?

Can you show in the PR an example output of a running dashboard?

tutorials/application_design/py/subscriber.py
```diff
@@ -45,71 +38,110 @@ def __init__(
...
+        dashboard_condition.set_handler(dashboard_handler)
+
+        display_handler_sentinel = threading.Event()
+        display_handler_condition = threading.Condition()
+
+        def display_handler():
+            while not display_handler_sentinel.is_set():
+                with display_handler_condition:
+                    display_handler_condition.wait(5)
+                    dashboard_condition.trigger_value = True
```
alexcamposruiz, line 68: can you explain why you use a GuardCondition instead of just calling `print(self._dashboard_string())` here?

tutorials/application_design/py/subscriber.py
```diff
@@ -45,71 +38,110 @@ def __init__(
...
-                print(
-                    f"  Known reached destinations: {len(data.reached_destinations)}"
+        dashboard_condition = dds.GuardCondition()
+
+        def dashboard_handler(_):
+            print(self._dashboard_string())
+
+        dashboard_condition.set_handler(dashboard_handler)
+
+        display_handler_sentinel = threading.Event()
```
alexcamposruiz, line 61: I'm not sure the event and the condition are necessary. You can just use `sleep` and rely on the `KeyboardException` to exit the sleep and the thread. I don't think you need to synchronize it by hand with an `Event`.

alexcamposruiz (commented):
Fix python formatting with black

alexcamposruiz (changes requested):
Suggested several improvements to reduce complexity and improve readability.

tutorials/application_design/py/subscriber.py
Alxe: I'm not sure if I'd call it useful, as it prints the pointer-like repr of the entities.

```
❯ python3 -m subscriber
Running dashboard: dashboard=Dashboard(self._metrics_reader=<rti.connextdds.DataReader object at 0x7f360e5e2bb0>, self._transit_reader=<rti.connextdds.DataReader object at 0x7f360e5324f0>)
[[ DASHBOARD: 2024-09-30 10:02:21.532963 ]]
Online vehicles: 0

Offline vehicles: 0

[INFO] Vehicle PALN6ZXN71WG318XU is enroute to Coord(lat=-46.10992285726171, lon=-23.36607730023902) from Coord(lat=-17.29690932734892, lon=36.59023701656847)

# ...

[[ DASHBOARD: 2024-09-30 10:02:26.533522 ]]
Online vehicles: 1
Vehicle PALN6ZXN71WG318XU:
  Fuel updates: 6
  Last known destination: Coord(lat=-46.10992285726171, lon=-23.36607730023902)
  Last known fuel level: 81.3764490923471

Offline vehicles: 0

# ...

[INFO] Vehicle PALN6ZXN71WG318XU has arrived at its destination in Coord(lat=29.106641859065363, lon=41.93699410392373)

[INFO] Vehicle PALN6ZXN71WG318XU is enroute to Coord(lat=9.017138169227357, lon=36.99741430219118) from Coord(lat=29.106641859065363, lon=41.93699410392373)

[INFO] Vehicle PALN6ZXN71WG318XU is enroute to Coord(lat=9.017138169227357, lon=36.99741430219118) from Coord(lat=-28.712438759150604, lon=32.36664575065243)

[[ DASHBOARD: 2024-09-30 10:02:46.534679 ]]
Online vehicles: 0

Offline vehicles: 1
Vehicle PALN6ZXN71WG318XU
```

tutorials/application_design/py/subscriber.py
Alxe: If I did not use a Guard Condition and instead printed after every dispatch, I would always get the dashboard printed whenever the waitset is dispatched.

Using a guard condition, I can make it so that the Guard Condition is printed consistely over time.

tutorials/application_design/py/subscriber.py
Alxe: I left them as local variables because they do not need to live within `self` and is much easier to explain step-by-step in the documentation.

tutorials/application_design/py/subscriber.py
```diff
@@ -45,71 +38,110 @@ def __init__(
...
+        except KeyboardInterrupt:
+            with display_handler_condition:
+                display_handler_sentinel.set()
+                display_handler_condition.notify_all()
+
+        display_thread.join()
+
+    def _new_position_string(self, condition):
+        string = str()
+        for sample, info in self._transit_reader.select().condition(condition).read():
```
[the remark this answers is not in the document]
Alxe: I am using `info` in line 91.

tutorials/application_design/py/subscriber.py
Alxe: Some type hints can be easily inferred, but creating an empty dictionary is not one of these scenarios. I'm limiting hinting to these scenarios instead of being explicit in every scenario.

tutorials/application_design/py/subscriber.py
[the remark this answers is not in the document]
alexcamposruiz: Checking `info.valid` is not necessary when you use `read_data()`; it already skips invalid data samples.

tutorials/application_design/py/subscriber.py
alexcamposruiz: I don't think that repr does anything useful then. Can you remove it or e.g. assign a name to the dashboard instead and print the name?

alexcamposruiz (changes requested):
Replied to open comments.

Also, make sure to apply relevant feedback to the C++ example as well.

tutorials/application_design/py/subscriber.py
```diff
@@ -30,86 +27,122 @@ class DashboardItem:
...
+            if not info.valid:
+                continue
 
-    async def run(self):
-        await asyncio.gather(
-            self._display_app(),
-            self._metrics_app(),
-            self._transit_app(),
+            string += f"[INFO] Vehicle {sample.vehicle_vin}"
+            if sample.current_route and len(sample.current_route):
```
alexcamposruiz, line 96:
```
if sample.current_route:
```
The and is always true.

alexcamposruiz (changes requested):

tutorials/application_design/py/subscriber.py
Alxe: I don't think this is accurate as `if sample.current_route` will evaluate to `sample.current_route.__bool__` which tests for whether if the optional is present or not. Once we've checked that, we need to check that the sequence is not empty.

alexcamposruiz: Both None and an empty list/string test to `False`:

```
>>> bool([])
False
>>> bool(None)
False
>>> bool([1])
True
```

alexcamposruiz (approved):

PR #711, opened 2024-09-23
Evalg 84

xFrenchy:
### Summary
Rewriting the EvalGuide Docker module to use Debian package for creating docker images with Connext installed

### Details and comments
-Updated Dockerfile to install packages such as curl, run the basic install commands for Connext
-Updated CmakeList to refer to /app to find cmake files
### Checks

-   [ x ] I have updated the documentation accordingly.
-   [ x ] I have read the [CONTRIBUTING](https://github.com/rticommunity/rticonnextdds-examples/blob/develop/CONTRIBUTING.md) document.


tutorials/docker/CMakeLists.txt
```diff
@@ -15,7 +15,8 @@ project(connext-tutorial-docker-pub-sub)
 
 # Include Connext CMake utilities
 list(APPEND CMAKE_MODULE_PATH
-    "/opt/rti.com/rticonnextdds-examples/resources/cmake/Modules/"
+    # "/opt/rti.com/rticonnextdds-examples/resources/cmake/Modules/"
```
alexcamposruiz, line 18: Delete the comment

tutorials/docker/Dockerfile
```diff
@@ -1,27 +1,59 @@
-FROM rticom/connext-sdk:7.3.0-EAR as build-stage
+ARG RTI_LICENSE_AGREEMENT_ACCEPTED=false
```
alexcamposruiz, line 1: why is this `ARG` defined here, before FROM and again in line 4?

alexcamposruiz, line 1: Instead of adding all this to this Dockerfile, my idea was to define a new Dockerfile, (e.g. `connext.Dockerfile`) that creates the `connext-sdk` image: 
```
docker build -t connext-sdk ./connext.Dockerfile`
```

(The git clone could be part of this image as well)

And then keep this Dockerfile very similar (starting with `FROM connext-sdk as build-stage`).

tutorials/docker/Dockerfile
```diff
@@ -1,27 +1,59 @@
...
+# Install the RTI Connext Debian Package
+RUN export DEBIAN_FRONTEND=noninteractive \
+    RTI_LICENSE_AGREEMENT_ACCEPTED=${RTI_LICENSE_AGREEMENT_ACCEPTED} \
+    && apt-get update \
+    && apt-get install -y rti-connext-dds-7.3.0 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists
 
 # For armv8 uncomment the following line
 # ENV CONNEXTDDS_ARCH=armv8Linux4gcc7.3.0
```
alexcamposruiz, line 36: It would be nice to figure out the right CONNEXTDDS_ARCH based on the system's archicteure (for ex. with `uname -m`)

tutorials/docker/Dockerfile
```diff
@@ -1,27 +1,59 @@
...
-COPY *.xml .
+COPY *.cxx ./
+COPY *.xml ./
+RUN git clone --recurse-submodule https://github.com/rticommunity/rticonnextdds-examples.git
 
 # Build the applications
 RUN mkdir build
 WORKDIR /app/build
-RUN cmake ..
+RUN cmake .. -DCONNEXTDDS_DIR=/opt/rti.com/rti_connext_dds-7.3.0/
```
alexcamposruiz, line 50: Is `-DCONNEXTDDS_DIR=` required? I think the cmake utils will look into that directory automatically.

alexcamposruiz (changes requested):

tutorials/docker/Dockerfile
xFrenchy: I got some information for how to design my dockerfile based on this: [https://bitbucket.rti.com/projects/PM/repos/tsn-arch-sim.git/browse/docker/Dockerfile]

I also spoke to Marco in the help-deb channel because previously I only had the ARG defined once inside the build-stage with a default value of false, and with my command line argument, the value was not changing. We thought that it needed to be defined with a default outside of the build-stage and then defined without a value inside the build stage to either inherit the command-line value or default value because previously it was only defined once in the build-stage and always using the default value. 

But later down the line I realized we were using the wrong value. It should have been "accepted" instead of "yes" or "true".

I just reran it by only declaring it once with a default in build-stage and it built fine. So I'll remove that

xFrenchy: I see, I'll decouple the connext-sdk building part into a different dockerfile and then try to link both of them back together

tutorials/docker/Dockerfile
xFrenchy: Without it, it was failing and asking me to specify it. If it continues failing without it, should we go ahead with -DCONNEXTDDS_DIR or look into why cmake util doesn't look into that directory automatically?

alexcamposruiz: it looks like `/opt/rti.com/...` is not a standard installation directory, so we need to specify the argument.

tutorials/docker/Dockerfile
xFrenchy: I haven't tested this on armv8, this comment was leftover from how the file was before. If everything else looks good, I'll test this part in the lab to see if setting CONNEXTDDS_ARCH is needed and if so, using uname -m or some other cpu command

PR #712, opened 2024-09-26
JH-76 -- Change version string to 7.4.0 for all cs/csproj files ...

alvinkrebs:
### Summary
These changes are required for the upcoming 7.4.0 release.

### Details and comments
The change for all csproj files is to use Version 7.4.0 instead of 7.3.0

**THIS MERGE SHOULD ONLY HAPPEN AFTER LAUNCH DATE**

### Checks


alexcamposruiz (approved):

elainerti (approved):

elainerti (changes requested):
You need to change the top level VERSION file to be 7.4.0 or else validation check will fail.

PR #717, opened 2024-10-03
updated readme for security plugins

akkshajs:
### Summary

Updated readme in connext_secure to include links to installation guide and current manual.

### Checks


colmenero (approved):

colmenero:
Wait for @lulivi to confirm the target branch.

We could discuss if we want to have these kind of links to always point to the latest documentation.
The advantage is that we don't have to manually update them. I can see that we have several outdated links in the examples.
The disadvantage is that we will have a documentation mismatch when users are running older versions of Connext. An alternative would be to update the links for every specific release-branch, but I don't know if this effort is worth it.

PR #723, opened 2024-10-31
#fix REQREPLY-221: Update RPC examples

albertorobles2000:
### Summary

With this PR we want to update RPC examples to use wait for service

### Details and comments

I have also updated the examples so they finish at some point, we need it to test them

### Checks


examples/connext_dds/remote_procedure_call/c++11/Inventory_service.cxx
```diff
@@ -81,31 +81,17 @@ class InventoryImpl : public InventoryService {
     std::mutex mutex;
 };
```
albertorobles2000, line 83: I created this function because we were creating the participnat in the same scope where we were calling `dds::domain::DomainParticipant::finalize_participant_factory();`. So the participant destructor was not invoked by the time the factory was finalized. producing a precondition error

examples/connext_dds/remote_procedure_call/c++11/Inventory_service.cxx
```diff
@@ -126,7 +112,38 @@ int main(int argc, char *argv[])
     ::InventoryServiceService service(service_impl, server, params);
 
     std::cout << "InventoryService running... " << std::endl;
-    server.run();
+    server.run(std::chrono::seconds(service_timeout));
```
albertorobles2000, line 115: For us to test the examples they have to run in a given time, they cannot run forever

alexcamposruiz:
The target branch can't be master; master contains code compatible with 7.3.0. You have to use develop for changes that will be supported in the next release. 

There's a caveat here though. The changes to `tutorials` (Development Guide) need to be compatible with 7.3.0 until the next LTS release, so we need to hold on to that for now. Since the changes are small, I think we could keep the previous wait, and add the new one in a comment:

```
# For versions 7.4.0 and below:
<current wait>
# For newer versions you can use the following:
# <new wait, commented>

alexcamposruiz (changes requested):

albertorobles2000:
The changes have been moved the [following PR](https://github.com/rticommunity/rticonnextdds-examples/pull/724)

PR #726, opened 2024-12-10
Property qos Python example

xFrenchy:
### Summary
Adding Python example for property_qos example

### Checks


xFrenchy:
Was delaying PR to add the interoperability test framework with it, however, considering that this example targets master, and the test will be in dev, I think I will split the PRs in 2 instead.

So this PR is ready for review now @Alxe

alexcamposruiz (changes requested):
I just realized I had feedback I hadn't submitted.

PR #741, opened 2025-07-08
ROUTING-1322 - Improving UDP socket adapter plugin example. Adding Data-diode reference in doc.

iutrilladn:
### Summary
Routing service adapter for udp socket example improvements.
- New example using dynamic data
- Old example is now bidirectional

### Details and comments
- udp_socket_adapter now includes converting from DDS to UDP socket
- New use case in README.md for data-diode environment.
- New example using dynamic data
- Split udp_socket example into two, dynamic and non dynamic example

about 99% of the code is the same in both examples.
I am open to reconfigure in a way that both can coexist in the same code using #ifdef statements, but I consider this way to be more clear.

Note:
Original example of udp_socket_adapter was not included in defaults for CMakeLists,txt. I have kept it the same way.

### Checks


examples/routing_service/udp_socket_adapter_dynamic/README.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+Then you can call CMake like this:
+
+```bash
+cmake -DCONNEXTDDS_DIR=<connext dir> -DCMAKE_TOOLCHAIN_FILE=<toolchain file created above>
+      -DCONNEXTDDS_ARCH=<connext architecture> ..
+```
+
+## Running C++ example
```
juanlu-rti, line 100:
```suggestion
## Running the C++ example
```

examples/routing_service/udp_socket_adapter_dynamic/src/SocketConnection.cxx
```diff
@@ -0,0 +1,73 @@
+/*
+ * (c) 2024 Copyright, Real-Time Innovations, Inc.  All rights reserved.
```
juanlu-rti, line 2:
```suggestion
 * (c) 2025 Copyright, Real-Time Innovations, Inc.  All rights reserved.
```

examples/routing_service/udp_socket_adapter_dynamic/src/SocketConnection.cxx
```diff
@@ -0,0 +1,73 @@
...
+void SocketConnection::delete_stream_reader(StreamReader *reader)
+{
+    SocketStreamReader *socket_reader =
+            dynamic_cast<SocketStreamReader *>(reader);
+    socket_reader->shutdown_socket_reader_thread();
+    delete reader;
+}
+void SocketConnection::delete_stream_writer(StreamWriter *writer)
+{
+    SocketStreamWriter *socket_writer = dynamic_cast<SocketStreamWriter *>(writer);
```
juanlu-rti, line 55: If the similar lines above needed wrapping then probably this one too

examples/routing_service/udp_socket_adapter_dynamic/src/SocketInputDiscoveryStreamReader.cxx
```diff
@@ -0,0 +1,88 @@
+/*
+ * (c) 2019 Copyright, Real-Time Innovations, Inc.  All rights reserved.
```
juanlu-rti, line 2:
```suggestion
 * (c) 2025 Copyright, Real-Time Innovations, Inc.  All rights reserved.
```

examples/routing_service/udp_socket_adapter_dynamic/src/SocketInputDiscoveryStreamReader.hpp
```diff
@@ -0,0 +1,61 @@
+/*
+ * (c) 2019 Copyright, Real-Time Innovations, Inc.  All rights reserved.
```
juanlu-rti, line 2:
```suggestion
 * (c) 2025 Copyright, Real-Time Innovations, Inc.  All rights reserved.
```

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamReader.cxx
```diff
@@ -0,0 +1,136 @@
+/*
+ * (c) 2019 Copyright, Real-Time Innovations, Inc.  All rights reserved.
```
juanlu-rti, line 2: Fix copyright if necessary, everywhere necessary

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamReader.cxx
```diff
@@ -0,0 +1,136 @@
...
+            continue;
+        }
+        /**
+         * Here we notify Routing Service, that there is data available
+         * on the StreamReader, triggering a call to take().
+         */
+
+         received_bytes_ = received_bytes;
+
+        reader_listener_->on_data_available(this);
```
juanlu-rti, line 64: Indenting seems off

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamReader.cxx
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    socket = std::unique_ptr<UdpSocket>(
+            new UdpSocket(receive_address_.c_str(), receive_port_));
+
+    socketreader_thread_ =
+            std::thread(&SocketStreamReader::socket_reading_thread, this);
+}
+
+void SocketStreamReader::take(
```
juanlu-rti, line 98: To increase throughput, perhaps we could do one or all of the following:
a) hold more than one set of received bytes to be able to batch more than 1 sample into `take()` operations.
b) hold an array of Dynamic Data samples instead of the bytes, already initialized.
  b1) and fill samples directly in the UDP receiving thread. That way we decouple deserialization from the critical path (`take()`).
  b2) we would also avoid memory allocations in the `take()` operation (because right now we create a new dynamic data object every time `take()` is called.
  b3) the number of Dynamic Data samples in the cache could be left to grow dynamically (2x per increase?) or could be initialized to `max_samples_per_read`. I'm not sure if you can access the value of this property when creating the Stream Reader. If you cannot you could perhaps set an XML property.
c) you can have the `std::vector<char>` you're using as a buffer for deserialising the samples in the class. That way it will grow dynamically as necessary and not with every call to `take()`.

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamReader.cxx
```diff
@@ -0,0 +1,136 @@
...
+
+using namespace dds::core::xtypes;
+using namespace rti::routing;
+using namespace rti::routing::adapter;
+
+void SocketStreamReader::socket_reading_thread()
+{
+    while (!stop_thread_) {
+        int received_bytes = 0;
+        socket->receive_data(
```
juanlu-rti, line 45: Is `receive_data` blocking? Could not receiving any more data create a race condition with `shutdown_socket_reader_thread`?

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamWriter.cxx
```diff
@@ -0,0 +1,123 @@
...
+            buffer.data(),
+            buffer.size(),
+            dest_address_.c_str(),
+            dest_port_);
+    }
+	
+    return len;
+}
+
+int SocketStreamWriter::write(
```
juanlu-rti, line 99: We're ignoring the `SelectorState`, which is fine, but then I believe we can skip implementing this method and leave just the base class version.

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamWriter.cxx
```diff
@@ -0,0 +1,123 @@
...
+        const std::vector<dds::core::xtypes::DynamicData *> &samples,
+        const std::vector<dds::sub::SampleInfo *> &infos,
+        const SelectorState &selector_state)
+{
+    int len;
+    len = write(samples, infos);
```
juanlu-rti, line 105: If you decide to keep this method, just do `return write(samples, infos);`

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamWriter.hpp
```diff
@@ -0,0 +1,77 @@
...
+    virtual int
+            write(const std::vector<dds::core::xtypes::DynamicData *> &,
+                  const std::vector<dds::sub::SampleInfo *> &) final;
+
+    virtual int write(
+            const std::vector<dds::core::xtypes::DynamicData *> &,
+            const std::vector<dds::sub::SampleInfo *> &,
+            const rti::routing::adapter::SelectorState &selector_state) final;
+
+    virtual void return_loan(
```
juanlu-rti, line 50: I don't think `StreamWriter` provides a `return_loan()` operation, in fact. Can you check? It's probably dead code.

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamWriter.hpp
```diff
@@ -0,0 +1,77 @@
...
+            SocketConnection *connection,
+            const rti::routing::StreamInfo &info,
+            const rti::routing::PropertySet &
+            );
+
+    virtual int
+            write(const std::vector<dds::core::xtypes::DynamicData *> &,
+                  const std::vector<dds::sub::SampleInfo *> &) final;
+
+    virtual int write(
```
juanlu-rti, line 45: Check also what I commented above about this `write` version with selector, I think you can just remove it.

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamWriter.hpp
```diff
@@ -0,0 +1,77 @@
...
+     * socket.
+     */
+    
+
+    SocketConnection *socket_connection_;
+
+    std::unique_ptr<UdpSocket> socket;
+
+    int send_port_;
+	int dest_port_;
```
juanlu-rti, line 69: Indenting is off.

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamWriter.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+    std::string send_address_;
+	std::string dest_address_;
```
juanlu-rti, line 72: Indenting is off

examples/routing_service/udp_socket_adapter_dynamic/src/UdpSocket.cxx
```diff
@@ -97,4 +97,15 @@ void UdpSocket::receive_data(
...
+}
+
+int UdpSocket::send_data(char* tx_buffer, int tx_length, const char* destAddr, int destPort)
+{
+    sockaddr_in dest_addr;
+    dest_addr.sin_family = AF_INET;
+	dest_addr.sin_port = htons(destPort);
+    dest_addr.sin_addr.s_addr = inet_addr(destAddr);
+
+    size_t length = sendto(sockfd, tx_buffer, tx_length, 0, (struct sockaddr*)&dest_addr, sizeof(dest_addr));
```
juanlu-rti, line 109: Could this block forever? Just a question. If the answer is yes, then let's add a timeout.

examples/routing_service/udp_socket_adapter_typed/README.md
```diff
@@ -5,6 +5,8 @@
 This example shows how to implement a simple Routing Service Adapter plugin
 in C++11 to receive data from a UDP socket using RTI Routing Service.
 
+This example requires including a Types.xml file with the data type information.
```
juanlu-rti, line 8: While we're using Dynamic Data for the outside endpoints, is it strictly necessary to provide an XML types file? What about when we use `route_types = true` in Routing Service? I believe it's the default now. Have you tested without XML types?

examples/routing_service/udp_socket_adapter_typed/RsSocketAdapter.xml
```diff
@@ -70,11 +69,61 @@
...
                 </route>
             </session>
         </domain_route>
     </routing_service>
 
-</dds>
+    <routing_service name="DDSToSocketAdapter">
+        <domain_route name="SocketBridge">
+            <connection name="SocketConnection" plugin_name="AdapterLib::SocketAdapter">
+                <registered_type name="ShapeType" type_name="ShapeType" />
```
juanlu-rti, line 82: Can you check if `register_type` is now the official tag to register a type? I believe `registered_type` may be deprecated. We don't want to use deprecated tags in examples.

examples/routing_service/udp_socket_adapter_dynamic/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+This examples uses dynamic data API and there is no need to know the data type
+information beforehand.
+
+The code in this directory provides the following components:
+
+-   `src/SocketAdapter` implements the plugin that is loaded by *RTI Routing
+Service*. It is responsible to create and delete connections.
+-   `src/SocketConnection` implements a connection. This component is
+responsible of the creation and deletion of `StreamReaders`.
```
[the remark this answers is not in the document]
iutrilladn: Done

examples/routing_service/udp_socket_adapter_dynamic/src/SocketInputDiscoveryStreamReader.cxx
iutrilladn: Done

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamReader.cxx
```diff
@@ -0,0 +1,136 @@
...
+        if (received_bytes <= 0) {
+            // Sleep for a small period of time to avoid busy waiting
+            std::this_thread::sleep_for(std::chrono::milliseconds(10));
+            continue;
+        }
+        /**
+         * Here we notify Routing Service, that there is data available
+         * on the StreamReader, triggering a call to take().
+         */
+
```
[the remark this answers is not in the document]
iutrilladn: Done

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamReader.cxx
```diff
@@ -0,0 +1,136 @@
...
+    socketreader_thread_ =
+            std::thread(&SocketStreamReader::socket_reading_thread, this);
+}
+
+void SocketStreamReader::take(
+        std::vector<dds::core::xtypes::DynamicData *> &samples,
+        std::vector<dds::sub::SampleInfo *> &infos)
+{
+    dds::core::xtypes::DynamicData deserialized_sample(*adapter_type_);
+    std::vector<char> received_buffer = std::vector<char>(received_buffer_, received_buffer_ + received_bytes_);
```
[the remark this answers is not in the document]
iutrilladn: Done

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamWriter.hpp
iutrilladn: Replaced \t with 4 spaces

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamWriter.cxx
```diff
@@ -0,0 +1,123 @@
...
+		send_port_));
+}
+
+int SocketStreamWriter::write(
+        const std::vector<dds::core::xtypes::DynamicData *> &samples,
+        const std::vector<dds::sub::SampleInfo *> &infos)        
+{
+    size_t len = 0;
+    for (const auto sample : samples) {
+        std::vector<char> buffer;
```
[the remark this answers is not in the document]
iutrilladn: Done

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamWriter.cxx
iutrilladn: Method deleted. Code not present anymore

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamWriter.hpp
iutrilladn: Deleted.
Correct, this is only necessary for the reader side and not writer.

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamReader.cxx
```diff
@@ -0,0 +1,136 @@
...
+            // Sleep for a small period of time to avoid busy waiting
+            std::this_thread::sleep_for(std::chrono::milliseconds(10));
+            continue;
+        }
+        /**
+         * Here we notify Routing Service, that there is data available
+         * on the StreamReader, triggering a call to take().
+         */
+
+         received_bytes_ = received_bytes;
```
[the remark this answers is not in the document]
iutrilladn: I have added a new queue to handle a vector of buffers.
Please check whether the approach is valid.

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamReader.cxx
iutrilladn: receive_data is not blocking. You can see its implementation at examples/routing_service/udp_socket_adapter_dynamic/src/UdpSocket.cxx

examples/routing_service/udp_socket_adapter_dynamic/src/UdpSocket.cxx
iutrilladn: The socket is configured non-blocking so it should be ok for both receive and send

`fcntl(sockfd, F_SETFL, O_NONBLOCK);`

examples/routing_service/udp_socket_adapter_dynamic/src/SocketStreamReader.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include "UdpSocket.hpp"
+
+#include <rti/routing/adapter/AdapterPlugin.hpp>
+#include <rti/routing/adapter/StreamReader.hpp>
+
+#define BUFFER_MAX_SIZE 1024
+#define RECEIVE_ADDRESS_STRING "receive_address"
+#define RECEIVE_PORT_STRING "receive_port"
+
+class SocketStreamReader : public rti::routing::adapter::DynamicDataStreamReader {
```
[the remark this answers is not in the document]
iutrilladn: Done

juanlu-rti:
Changes applying feedback look fine. Did you measure performance after the changes? Just out of curiosity. Anyhow, approving.

juanlu-rti (approved):

iutrilladn:
Hi @juanlu-rti 
Thanks for the review. I did not measure performance after the changes.
I don't think I have write permissions. Could you merge for me of give me write permission for this merge?

Issue #745, opened 2025-10-08
[BUG] Fix details on Recording Service examples


Antoniolm:
### System information

- **RTI Product**: Rti Recording Service
- **Version**: 7.3.1
- **Operating system**: Windows
- **Compiler**: Visual Studio
- **Compiler version**: 2017
- **Additional information**:

### What is the current behavior?
- Warning building service admin example
``` 
...examples\recording_service\service_admin\c++11\requester.cxx(240): warning C4244: '=': conversion from 'uint64_t' to 'uint32_t', possible loss of data [...\examples\recording_service\service_admin\c++11\build
```
- Issues on pluggable storage 
  * Cpp_PluggableStorage.dat.info file may not contain the finish timestamp, just the starting one. 
  * There is a third generated file Cpp_PluggableStorage.dat.pub which is not mentioned in the README:
  * The print format of the subscriber seems to be wrong on Windows
```
]id: 50, msg:  Sample string 50
```

PR #746, opened 2025-10-08
#745 Fix Recording Service issues on several modern C++ examples #RECORD-1555 #RECORD-1553

Antoniolm:
### Checks


examples/recording_service/pluggable_storage/c++11/FileStorageWriter.cxx
```diff
@@ -70,16 +70,22 @@ FileStorageWriter::FileStorageWriter(
...
-        info_file_ << "End timestamp: " << current_time << std::endl;
+    if (info_file_.fail()) {
+        std::cerr << "Failed to use file to store metadata";
+    }
+
+    /* Obtain current time */
+    int64_t current_time = (int64_t) time(NULL);
+    if (current_time == -1) {
+        // can't throw in a destructor
+        std::cerr << "Failed to obtain the current time";
```
juanlu-rti, line 82: Should we be returning here? It does not make sense to use `current_time` after checking it's -1

examples/recording_service/pluggable_storage/c++11/FileStorageWriter.cxx
```diff
@@ -70,16 +70,22 @@ FileStorageWriter::FileStorageWriter(
...
-        /* Obtain current time */
-        int64_t current_time = (int64_t) time(NULL);
-        if (current_time == -1) {
-            // can't throw in a destructor
-            std::cerr << "Failed to obtain the current time";
-        }
-        /* Time was returned in seconds. Transform to nanoseconds */
-        current_time *= NANOSECS_PER_SEC;
-        info_file_ << "End timestamp: " << current_time << std::endl;
+    if (info_file_.fail()) {
```
juanlu-rti, line 73: What could cause the `info_file_` to be in a bad state?

Antoniolm: Using the same approach that we do in the constructor function. The problem that we found it related to the end timestamp being missing of the file. With this instrumentation, we can know at which point it is failing.

examples/recording_service/pluggable_storage/c++11/FileStorageWriter.cxx
Antoniolm: Applying the approach

juanlu-rti (approved):

